PR #5, opened 2020-10-11
Improved Menu Items and added contents for DS Menu

sumitd94:


sumitd94:
Adding a new PR, Messed up with file naming conventions

akankshadharkar:
I just saw this 

A couple of things - 
1. Do you think NavBar would be a better name than Menu?
2. Can we add screenshots of the changes for reviews from next PR onwards?

PR #18, opened 2020-12-16
 Converted all class-based into functional components

Khanjan04:
Converted a class-based component into functional component.

### DIFF
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -1,5 +1,4 @@
 import styles from "./Layout.module.scss";
-import React from 'react';
 
 type Props = {
   children?: React.ReactNode;
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -1,5 +1,4 @@
 import styles from "./Layout.module.scss";
-import React from "react";
 
 type Props = {
   children?: React.ReactNode;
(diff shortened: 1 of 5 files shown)


src/components/title/index.tsx
```diff
@@ -1,4 +1,5 @@
 import styles from "./title.module.scss";
+import React from "react";
```
sumitd94, line 2: NIT: Importing react is not necessary. In `Next JS` by default we have react imported

src/components/Layout/index.tsx
```diff
@@ -1,4 +1,5 @@
 import styles from "./Layout.module.scss";
+import React from 'react';
```
sumitd94, line 2: NIT: Importing react is not necessary. In Next JS by default we have react imported

sumitd94 (changes requested):

src/components/title/index.tsx
Khanjan04: Okay. Actually it is showing some error otherwise. Hence I imported react.

Khanjan04:
I have made the required changes. Please have a look.

src/components/title/index.tsx
swarajpure: @Khanjan04 After the new commit, does the code work properly? Because you said it was showing error if react was not imported.

sumitd94 (dismissed):

sumitd94 (approved):

whyDontI (approved):
Great work @Khanjan04 👍🏽 🥳

PR #34, opened 2021-01-01
Github actions 

### LINKED ISSUE #33: Set up Github Actions
Github actions need to be set up

Why is it needed?
- If `npm run build` fails for a particular PR/commit, we will get notified that the build is failing for this one. So that we can avoid merging it.
- If such PR gets merged, the build will fail on production which is not desirable.
- Knowing problems beforehand will help us fix them earlier.

What to do?
- run `npm ci` (why not `npm install`? refer this one: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-nodejs)
- run `npm run build` 
- TODO: run for tests once we have them in place

Refer: https://github.com/Real-Dev-Squad/website-crypto/blob/develop/.github/workflows/node.js.yml

agarwalamn:
# Closes #33 
Added node.js.yml file for issue #34 

As we are using yarn to build things used yarn commands


.github/workflows/node.js.yml
```diff
@@ -0,0 +1,28 @@
...
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: yarn
```
swarajpure, line 27: Not `yarn install`?

swarajpure (commented):
Thank you!

I have a doubt. Please clarify.

.github/workflows/node.js.yml
agarwalamn: yarn and yarn install works the same

swarajpure: Cool.

Also, can the name of file be self-explanatory?

Something like `continuous-integration`?

agarwalamn: sure, I thought we have node.js as the default schema for the file name😅

agarwalamn:
@swarajpure renamed the file

swarajpure (approved):
Thank you! 😄

Issue #48, opened 2021-01-16
Create useFetch custom hook to make API calls


whyDontI:
## Reference
https://reactjs.org/docs/hooks-custom.html

## Usage
```
const { response, loading, error } = useFetch(url);
```

PR #61, opened 2021-01-20
Add useFetch custom hook

Cartmanishere:
Added a custom hook to make API calls called `useFetch` as given in https://github.com/Real-Dev-Squad/website-status/issues/48


whyDontI (approved):
Great work 🥳 👍🏽

PR #65, opened 2021-01-28
Use useFetch hook in tasks page

### LINKED ISSUE #64: Use useFetch hook in tasks page

whyDontI:
Closes #64 

- [x] Refactored useFetch
- [x] Used useFetch in tasks page


sumitd94 (dismissed):
Looks good to me.

src/hooks/useFetch.ts
```diff
@@ -1,22 +1,29 @@
-import React from 'react';
+import { useState, useEffect } from 'react';
+import fetch from '@/helperFunctions/fetch';
 
-const useFetch = (url: string, options: object = {}) => {
-  const [response, setResponse] = React.useState(null);
-  const [error, setError] = React.useState(null);
-  const [isLoading, setIsLoading] = React.useState(false);
-  React.useEffect(() => {
+const useFetch = (url: string) => {
```
Cartmanishere, line 4: If this is a reusable utility, I think a way to provide options would be good to have. So it can be used for any type of usecase.

src/hooks/useFetch.ts
```diff
@@ -1,22 +1,29 @@
...
-        const json = await res.json();
-        setResponse(json);
+        const res = await fetch({
+          url,
+          method: 'get',
+        });
+        setResponse(res.data);
         setIsLoading(false);
       } catch (err) {
+        setIsLoading(false);
```
Cartmanishere, line 20: finally block will run regardless of the error not. So this is not needed here.

Cartmanishere (approved):

PR #69, opened 2021-02-06
Update continuous-integration.yml

sanyogita14:
Updated the node-versions in continuous-integration.yml file


whyDontI (approved):
Thanks 👍🏽

PR #78, opened 2021-02-23
Use built-ins, updated API routes

correaswebert:
* `next/head` provides the functionalities provided by `React Helmet`.
  So no need for an additional dependency.
* The favicon was named `RDSLogo.png`, so renamed it from the
  default of `favicon.ico`.

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,2 @@
-NEXT_PUBLIC_BASE_URL='https://staging-api.realdevsquad.com'
+NEXT_PUBLIC_BASE_URL='https://api.realdevsquad.com'
 NEXT_PUBLIC_GITHUB_IMAGE_URL='https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members/'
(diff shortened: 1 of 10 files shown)


.env
```diff
@@ -1,2 +1,2 @@
-NEXT_PUBLIC_BASE_URL='https://staging-api.realdevsquad.com'
+NEXT_PUBLIC_BASE_URL='https://api.realdevsquad.com'
```
whyDontI, line 1: Backend team will be hosting staging API soon, we don't need to change all the env files.
We want the development environment to use staging API.

We can revert this change.

.env.development
```diff
@@ -1,2 +1,2 @@
-NEXT_PUBLIC_BASE_URL='https://staging-api.realdevsquad.com'
+NEXT_PUBLIC_BASE_URL='https://api.realdevsquad.com'
```
whyDontI, line 1: Not needed as explained in previous comment

whyDontI (dismissed):
Good catch 👍🏽 
Write a good description for this PR, explaining the changes you did and the reason.
Please create an issue and mention the issue in the description of this PR.

Requested some little changes, everything else looks good 👍🏽 🥳

swarajpure (commented):
If we're not at all going to use `react-helmet`, can we remove that package from the `package.json?`

.env
correaswebert: Okay! Will revert 👍🏾

correaswebert:
> If we're not at all going to use `react-helmet`, can we remove that package from the `package.json?`

whyDontI (approved):
Good work @correaswebert 👍🏽

Issue #73, opened 2021-02-20
Show active tasks in Active section


whyDontI:
## Problem
- Active tasks are being shown in the completed section
![Screenshot from 2021-02-20 22-21-14](https://cdn.discordapp.com/attachments/673083527624917003/814387823998271488/unknown.png)

PR #81, opened 2021-03-10
Fixed issue with filtering task based on status

ice-rahul:
[issue 73](https://github.com/Real-Dev-Squad/website-status/issues/73) 

issue_73

PR to fix the status issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "husky": "^4.3.8",
+    "husky": "^5.0.9",
     "lint-staged": "^10.5.3"
   },
   "engines": {
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -40,7 +40,6 @@ const Index: FC = () => {
 
       <div className={classNames.container}>
         {!!error && <p>Something went wrong, please contact admin!</p>}
-        {(!filteredTask.length && !isLoading) && <p>No tasks found</p>}
         {
           isLoading
             ? (
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -38,14 +38,15 @@ const Index: FC = () => {
     <Layout>
       <Head title="Tasks" />
 
-      <div className="container">
+      <div className={classNames.container}>
         {!!error && <p>Something went wrong, please contact admin!</p>}
+        {(!filteredTask.length && !isLoading) && <p>No tasks found</p>}
         {
           isLoading
             ? (
               <p>Loading...</p>
             ) : (
-              <div className={classNames.container}>
+              <>
                 {
                   Object.keys(filteredTask).map((key) => (
                     <Accordion title={key}>
@@ -53,7 +54,7 @@ const Index: FC = () => {
                     </Accordion>
                   ))
                 }
-              </div>
+              </>
             )
         }
       </div>
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -9,33 +9,30 @@ import Accordion from '@/components/Accordion';
 
 const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
 
-const renderCardList = (tasks: task[]) => tasks.map(
-  (item: task) => <Card content={item} key={item.id} />,
-);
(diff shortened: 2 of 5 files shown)


swarajpure (dismissed):
Less time complexity, less hardcoding, removed unnecessary re-renders.

Great job 🥳 
Approving from my side.

swarajpure:
Congratulations on your first PR in open source and that too in Real Dev Squad 🥳 🎉

ice-rahul:
> Congratulations on your first PR in open source and that too in Real Dev Squad 🥳 🎉

thanks :)

harshith-venkatesh (dismissed):

src/pages/index.tsx
```diff
@@ -9,33 +9,28 @@ import Accordion from '@/components/Accordion';
 
 const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
 
-const renderCardList = (tasks: task[]) => tasks.map(
-  (item: task) => <Card content={item} key={item.id} />,
-);
+function renderCardList(tasks: task[]) {
```
harshith-venkatesh, line 12: Any reason for shifting away from arrow function approach

harshith-venkatesh: Rest all LGTM

package.json
```diff
@@ -45,7 +45,7 @@
     "eslint-plugin-prettier": "^3.3.1",
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "husky": "^4.3.8",
+    "husky": "^5.0.9",
```
whyDontI, line 48: Let's not push the code which is not required for the current issue we are solving.
But as this is a small change and not affecting much, we can keep this change for now.

Let's keep this in mind from future 😃

whyDontI (changes requested):
Nice work 🎉 
Just we need to handle the case when there are no tasks in the response.
Show some error message on the page.

Rest all looks good 👍🏽

src/pages/index.tsx
ice-rahul: If we would be doing it the other way around the function expression will become a one-liner as per ESLINT standards.
Also, that one-liner will be of length 106 characters. But, ESLINT allows only 100 characters in a line
So changed it to function declaration.

package.json
ice-rahul: Actually, I was facing an issue while committing it.
So in order to fix the commit issue, I had to update it

src/pages/index.tsx
harshith-venkatesh: Hmmmm, I mean changing the way just for eslint issue is something really new to me.
But ok, you should have thought a lot before coming to this conclusion so cool

harshith-venkatesh (dismissed):

package.json
whyDontI: What issue, can you please explain here?

ice-rahul: it was related to the pre-commit hook

**pre-commit hook failed (add --no-verify to bypass)**

[https://github.com/typicode/husky/issues/716](https://github.com/typicode/husky/issues/716)

src/pages/index.tsx
```diff
@@ -44,24 +39,21 @@ const Index: FC = () => {
...
-                <Accordion title="Active">
-                  {renderCardList(activeTasks)}
-                </Accordion>
-                <Accordion title="Completed">
-                  {renderCardList(completeTasks)}
-                </Accordion>
+                {
+                  Object.keys(filteredTask).length > 0
+                    ? Object.keys(filteredTask).map((key) => (
+                      <Accordion title={key}>
```
sumitd94, line 52: `key` is missing here, can we please add it.

sumitd94 (changes requested):
One small change

src/pages/index.tsx
ice-rahul: Do we really need to add it?
As ESLINT is not having issues also we are having a unique title prop attached to Accordion.

whyDontI: https://reactjs.org/docs/lists-and-keys.html

whyDontI (approved):
Nice work 👍🏽

src/pages/index.tsx
ice-rahul: > https://reactjs.org/docs/lists-and-keys.html
> @ice-rahul You might want to go through this.

thanks for the reference 👍

PR #89, opened 2021-03-19
Completed Mine Task

saxenaaman628:
Completed Mine Task


src/pages/mine.tsx
```diff
@@ -1,11 +1,58 @@
...
+import useFetch from '@/hooks/useFetch';
+import classNames from '@/styles/tasks.module.scss';
+import { task } from '@/components/constants/types';
 
-const Mine: FC = () => (
+const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks/self`;
+
+
+const CardList = (tasks: task[]) => tasks.map(
+    (item: task) => <Card content={item} key={item.id} />);
```
correaswebert, line 13: we could shift the arrow fn body to the next line OR
put the closing paranthesis on the next line

src/pages/mine.tsx
```diff
@@ -1,11 +1,58 @@
...
+    if ('tasks' in response) {
+      setTasks(response.tasks);
+
+    }
+  }, [isLoading, response]);
+  return(
   <Layout>
     <Head title="Mine" />
+    <div className={classNames.container}>
+        {!!error && <p>Something went wrong, please contact admin!</p>}
```
correaswebert, line 33: both the lines are shown...
```
Something went wrong, please contact admin!
No Tasks Found
```

Please show only error in case of error

src/pages/mine.tsx
```diff
@@ -1,11 +1,58 @@
...
+        {
+          isLoading
+            ? (
+              <p>Loading...</p>
+            ) : (
+
+              <>
+                {
+                  Object.keys(tasks).length > 0
```
correaswebert, line 42: I guess just `tasks.length` should suffice

src/pages/mine.tsx
```diff
@@ -1,11 +1,58 @@
...
+                    ?  (
+                      <div>
+                        {CardList(tasks)}
+                      </div>
+                    ) : 'No Tasks Found'
+                }
+              </>
+            )
+        }
+      </div>
```
correaswebert, line 52: There is an extreme amount of white spaces
You could reduce them to something like...
```js
<div className={classNames.container}>
  {!!error && <p>Something went wrong, please contact admin!</p>}
  
  {isLoading ? (
    <p>Loading...</p>
  ) : (
    tasks.length > 0 ? <div>{CardList(tasks)}</div> : "No Tasks Found"
  )}
</div>
```

src/components/tasks/mine_card/index.tsx
```diff
@@ -0,0 +1,33 @@
+import { FC } from 'react';
+import classNames from '@/components/tasks/mine_card/card.module.scss';
+import { task } from '@/components/constants/types';
+
+type Props = {
+  content: task,
+};
+
+const informationElement = (title: string, value: string) => (
```
correaswebert, line 9: You could convert this into a component instead of a function

correaswebert (changes requested):
Please use some formatter like Prettier so that you won't have to worry about formatting 😄

swarajpure:
> Please use some formatter like Prettier so that you won't have to worry about formatting smile

Better to use ESLint (Formatter + Linter) 😉

whyDontI:
> Please use some formatter like Prettier so that you won't have to worry about formatting

https://github.com/Real-Dev-Squad/website-status/blob/develop/.eslintrc.json

saxenaaman628:
Please review it

src/components/tasks/mine_card/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <div className={classNames.card}>
+      <span className={classNames.prTitle}>{title}</span>
+      {informationElement('Started', startedOn)}
+      {informationElement('Status', status)}
+    </div>
+  );
+};
+
+export default Card;
```
whyDontI, line 32: Any specific reason for creating a new card component separate from mine tasks when we already have a card component?
If not, we can just use `tasks/card` instead of creating this new component.

src/pages/mine.tsx
```diff
@@ -1,11 +1,54 @@
-import { FC } from 'react';
+import { FC, useState, useEffect } from 'react';
 import Head from '@/components/head';
 import Layout from '@/components/Layout';
+import Card from '@/components/tasks/mine_card';
```
whyDontI, line 4: Seems like we don't need the mine_card component as described in previous comment
`import Card from '@/components/tasks/card';`

whyDontI (changes requested):
Good work, requesting some changes. Please address 😃

src/components/tasks/mine_card/index.tsx
saxenaaman628: I thought Like for mine, we need to see less info but I will change it to card only

whyDontI (approved):
Great work @saxenaaman628 🥳

PR #91, opened 2021-03-22
[feature] Added pagination to PRs

### LINKED ISSUE #77: Create pagination
Only display a certain number of issues, and paginate their display.

Currently, only the first 10 or so issues are queried, and if the limit is increased, thy will be displayed at once.
Instead, created a paginated view, so that the response payload is small, and UI is convinient.

correaswebert:
Closes #77


src/pages/openPRs.tsx
```diff
@@ -51,26 +62,55 @@ const openPRs: FC = () => {
...
+          margin: 1rem;
+          padding: 1rem;
+          font-size: 1.5em;
+          cursor: pointer;
+          border: none;
+          background-color: #bfbfbf;
+          border-radius: 5px;
+        }
+      `}
+      </style>
```
whyDontI, line 113: Let's put the styles in a separate files.

whyDontI (changes requested):
Let's create a separate component and use the same across the pages, let's not repeat the same code on all the pages.

src/pages/openPRs.tsx
correaswebert: Yeah sure 👍🏾

src/pages/stale-pr.tsx
```diff
@@ -1,30 +1,117 @@
...
+
+        <button
+          className="pagination-btn"
+          type="button"
+          onClick={handleNext}
+          disabled={pullRequests.length === 0}
+        >
+          Next
+        </button>
+      </div>
```
sumitd94, line 89: The similar code we have on `pages/openPRs.tsx`
why not refactor this and add it in a component and call that component at both pages?

src/pages/stale-pr.tsx
```diff
@@ -1,30 +1,117 @@
...
+          margin: 1rem;
+          padding: 1rem;
+          font-size: 1.5em;
+          cursor: pointer;
+          border: none;
+          background-color: #bfbfbf;
+          border-radius: 5px;
+        }
+      `}
+      </style>
```
sumitd94, line 113: Can we please move this css in a separate css file.

sumitd94 (changes requested):
Request few changes.

src/components/PullRequestList/PullRequestList.module.scss
```diff
@@ -0,0 +1,19 @@
+.center_text {
+    font-size: 1.5em;
+    text-align: center;
+  }
+
+.pagination {
+display: flex;
+justify-content: center;
```
sumitd94, line 8: can we please fix the tab indentation.

src/components/PullRequestList/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const prUrl = `${process.env.NEXT_PUBLIC_BASE_URL}/pullrequests/${prType}?page=${page}`;
+
+  const fetchPRs = async () => {
+    try {
+      setLoading(true);
+      const response = await axios.get(prUrl);
+      setPullRequests(response.data.pullRequests);
+
+      if (!response.data.pullRequests.length) {
```
sumitd94, line 35: NIT: For readability purpose, you can destructure the `response.data` like this
```javascript
const { pullRequests } = response.data
```
And then use this `pullRequests` across.
Will make it more readable

sumitd94 (changes requested):
Thanks for the changes, good job 🥇 
Just one small change and then its good to go.

src/components/PullRequestList/index.tsx
correaswebert: I wanted to just set it and use the state of pullRequests instead
but somehow the state was not changing synchronously in my machine 😓 

Otherwise the if condition will check on the state pullRequests instead of the property buried inside the response.
Please test if it works on your machine.

src/components/PullRequestList/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        createdAt={createdAt}
+        updatedAt={updatedAt}
+        url={link}
+      />
+    );
+  });
+
+  return (
+    <Layout>
+      <Head title="Stale PRs" />
```
whyDontI, line 78: Title should be configurable

whyDontI (changes requested):
Great work 😊, requesting a little change.
Also, address other review comments or follow-up.

whyDontI (changes requested):
Pagination is not working for Stale PRs
Can you check?

Issue #95, opened 2021-03-26
Remove border from task card


ankushdharkar:
We would not be wanting this black border to appear anymore:

![image](https://user-images.githubusercontent.com/1935403/112640931-17c69800-8dff-11eb-987b-98aafbd2ecff.png)

sebuelias:
I would like to take this issue.
But on removing the border it might look a little bad, so any suggestions on what to do after? Or leave It as is?

swarajpure:
> I would like to take this issue.
> But on removing the border it might look a little bad, so any suggestions on what to do after? Or leave It as is?

One idea I can think of is, to have a look at the member cards on `members.realdevsquad.com`

sebuelias:
Is this look fine? If not, please give inputs.
![image](https://user-images.githubusercontent.com/10326494/113450066-cc267680-941c-11eb-8be1-5d72d83b98a6.png)

Or do I have to make any other changes?

Also if I have some suggestion as to may be modify in the same component, I need to raise a new issue and get it approved, right?

swarajpure:
> Is this look fine? If not, please give inputs.
> ![image](https://user-images.githubusercontent.com/10326494/113450066-cc267680-941c-11eb-8be1-5d72d83b98a6.png)
> 
> Or do I have to make any other changes?
> 
> Also if I have some suggestion as to may be modify in the same component, I need to raise a new issue and get it approved, right?

Looks good to me. You can make a PR. Make sure to add screenshots of before and after so the reviewers know what's done.

PR #103, opened 2021-04-03
issue/95- box-shadow added for task card

sebuelias:
fixes:  #95 

Before : 
image

After:
image

### DIFF
--- a/src/components/tasks/card/card.module.scss
+++ b/src/components/tasks/card/card.module.scss
@@ -5,8 +5,12 @@
   flex-wrap: wrap;
   margin-bottom: 2em;
   cursor: pointer;
-  box-shadow: 0 0 15px -7px rgb(0, 0, 0 / 65%);
+  box-shadow: 0 0 15px -7px rgba(0, 0, 0, 0.65);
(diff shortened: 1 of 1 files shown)


src/components/tasks/card/card.module.scss
```diff
@@ -5,6 +5,8 @@
   flex-wrap: wrap;
   margin-bottom: 2em;
   cursor: pointer;
+  box-shadow: 0 0 15px -7px rgb(0, 0, 0 / 65%);
+  border-radius: 10px;
```
swarajpure, line 9: What about hovering effect?

sebuelias: My bad. Will fix that as well.

sebuelias: image

updated the same as above, is that ok?

swarajpure: Yes, please push it

swarajpure (approved):
Looks good to me!

Thanks for this one! 🎉 

Congratulations on your first pull request in Real Dev Squad! 😄

Issue #102, opened 2021-04-03
Remove the extra round bracket


RohitKuwar:
Remove the extra unnecessary round bracket present on the [OpenPRs](https://status.realdevsquad.com/openPRs) page.

![Screenshot (576)](https://user-images.githubusercontent.com/70262783/113472833-4a1d6880-9483-11eb-93d3-035c38439e37.png)

nilisha-jais:
I want to take it up

PR #104, opened 2021-04-03
Removed the extra round bracket

nilisha-jais:
#102 has been solved


swarajpure (approved):
Good one. Thank you! 🎉

PR #111, opened 2021-04-10
issue/93 - Accordion width updated and centered in the parent

### LINKED ISSUE #93: Width on desktop 750px for Accordion Title
image

sebuelias:
parent - task_container.

fix #93 

On changing the maximum width to 750px it will look like this : 
image

After aligning to center it looks better, so updating with the same
image

Do comment with the same incase of any issues


whyDontI (approved):
Great work 🥳

PR #118, opened 2021-04-13
Dummy profile picture and no contributor text

whyDontI:
# Preview

Screenshot from 2021-04-13 19-05-26


src/components/tasks/card/index.tsx
```diff
@@ -20,11 +20,12 @@ const Card: FC<Props> = ({ content }) => {
     title,
     endsOn,
     startedOn,
-    ownerId,
     status,
+    assignee,
   } = content;
 
-  const ownerProfilePic = `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}${ownerId}/img.png`;
+  const [assigneeProfilePic, setAssigneeProfilePic] = useState(`${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}${assignee}/img.png`);
```
sumitd94, line 27: NIT: just for readability, can we store the value under `useState` in a variable and then use that variable here?

sumitd94 (approved):
One NIT Comment, but that won't stop this from being approved.

harshith-venkatesh (approved):

src/components/tasks/card/index.tsx
whyDontI: Yes, we can,
But as it is already `assigneeProfilePic`, and we will need a similar name for the value, which can be confusing.
Let's handle this later 😃 

Thanks for the review 😄

PR #139, opened 2021-05-26
Infinite scroll for openpr

### LINKED ISSUE #44: Load more open PRs upon reaching the end
When a user scrolls down to the last row of the N items, show the next set of N items

image

saxenaaman628:
Closes #44


whyDontI (changes requested):
Good work @saxenaaman628 👍🏽 
Requesting some changes,
- Can we not have the extra scrollbar
- When the user hits the end of the screen, can we not load the whole page.

Also, I can see you have made some UI changes that were not requested or that are not related to the issue. Can you please revert the same?
- OpenPRs cards should be in a Grid as before
- Update the Skeleton UI loader to match with the current style of the card

ScrollAnimation

src/components/PullRequestList/index.tsx
```diff
@@ -1,11 +1,11 @@
-import { FC, useState, useEffect } from 'react';
+import React, { FC, useState, useEffect } from 'react';
 import Head from '@/components/head';
 import Layout from '@/components/Layout';
 import PullRequest from '@/components/pullRequests';
-import CardShimmer from '@/components/Loaders/cardShimmer';
 import axios from 'axios';
 import styles from './PullRequestList.module.scss';
 
+axios.defaults.timeout = 3000;
```
whyDontI, line 8: Use our custom hook for making requests.
https://github.com/Real-Dev-Squad/website-status/blob/develop/src/hooks/useFetch.ts

src/components/PullRequestList/index.tsx
```diff
@@ -1,11 +1,11 @@
-import { FC, useState, useEffect } from 'react';
+import React, { FC, useState, useEffect } from 'react';
```
whyDontI, line 1: Next.js doesn't need `React` import I guess, Please confirm and do the needful.

src/components/PullRequestList/index.tsx
```diff
@@ -20,88 +20,83 @@ type PullRequestListProps = {
...
+        || document.body.scrollTop;
+      const scrollHeight = (document.documentElement
+          && document.documentElement.scrollHeight)
+          || document.body.scrollHeight;
+      if (scrollTop + window.innerHeight + 50 >= scrollHeight) {
+        setIsBottom(true);
+      }
+    };
+    window.addEventListener('scroll', onScroll);
+    return () => window.removeEventListener('scroll', onScroll);
```
whyDontI, line 73: Why two?

whyDontI (changes requested):

src/components/PullRequestList/index.tsx
saxenaaman628: Removing the event handler to prevent any memory leak. If somebody using older browser.

src/components/PullRequestList/index.tsx
saxenaaman628: Used axios instead of custom fetch because there are multiple hooks to set empty array in case of error and detecting bottom.
Was not able to do the same with custom hooks.

src/components/PullRequestList/index.tsx
saxenaaman628: yeah we dont need it.

src/components/PullRequestList/index.tsx
whyDontI: Oh my bad, I thought it was 2 `addEventListeners` 😶

src/components/PullRequestList/index.tsx
whyDontI: Is there any way you can make the custom hook usable in this usecase?
If yes let me know, we'll create an issue for making it more configurable.

saxenaaman628: Yes, I think I can make it using custom fetch and A new issue would be better,then doing it in this PR.

whyDontI: Can you create the issue, explain what changes you want.

saxenaaman628: Ok, So this PR will be merged?

saxenaaman628: created new issue.Please check

whyDontI: Can you make use of the `useFetch` hook as shown in the below example?
```js
  const {
    response,
    error,
    isLoading,
  } = useFetch(OPEN_PULL_REQUEST_URL);
  
  useEffect(() => {
    // Handle error
    // Empty array
  }, [error])
```

src/components/PullRequestList/index.tsx
```diff
@@ -18,90 +17,87 @@ type PullRequestListProps = {
...
-    setPage(page > 1 ? page - 1 : 1);
-  };
-
-  const handleNext = () => {
-    setPage(pullRequests.length ? page + 1 : page);
-  };
+    if (response?.pullRequests?.length > 0) {
+      if (page > 1) {
+        setPullRequests((pullRequest) => [...pullRequest, ...response.pullRequests]);
+      } else { setPullRequests(response.pullRequests); }
```
correaswebert, line 53: Please indent the `else` block consistently 😄

saxenaaman628: No we don't I will change it, Thanks

src/components/PullRequestList/PullRequestList.module.scss
```diff
@@ -17,3 +17,9 @@
   background-color: #bfbfbf;
   border-radius: 5px;
 }
+.scroll {
+  display: flex;
+	flex-direction: column;
+	align-items: center;
+  width: 100%;
```
swarajpure, line 24: Please fix the indentation here

saxenaaman628: Done

src/components/PullRequestList/index.tsx
```diff
@@ -18,90 +17,85 @@ type PullRequestListProps = {
...
+                title={title}
+                username={username}
+                createdAt={createdAt}
+                updatedAt={updatedAt}
+                url={link}
+              />
+            );
+          })}
+        </ul>
+        {isLoading
```
correaswebert, line 96: `isLoading && <html>` when the else is not required 😄

src/components/PullRequestList/index.tsx
```diff
@@ -18,90 +17,85 @@ type PullRequestListProps = {
...
-        <button
-          className={styles.pagination_btn}
-          type="button"
-          onClick={handleNext}
-          disabled={pullRequests.length === 0}
-        >
-          Next
-        </button>
+      <div className={styles.scroll}>
+        { error ? <p className={styles.center_text}>Something went wrong! Please contact admin</p> : ''}
```
correaswebert, line 78: `error && <html>` when else is not required 😄

correaswebert (commented):
Some changes, not necessary, but will shorten code 👍🏾

whyDontI:
I think I am missing something because there is no way we can use `useFetch` hook to make requests on the fly.
Hooks only execute at the start.
We need to use our custom fetch function.
Please let me know if I am missing something

src/components/PullRequestList/index.tsx
whyDontI: > Please ignore above suggestion, I think it won't work for making requests without a refresh / remount

We can use the `useFetch` to make a request only at the mounting phase of the component.
You'll need to use our custom `fetch` function.
Let me know if I am missing something.

correaswebert:
> I think I am missing something because there is no way we can use `useFetch` hook to make requests on the fly.
> Hooks only execute at the start.
> We need to use our custom fetch function.
> Please let me know if I am missing something

It's using a `useEffect` under the hood, with the url as dependency.
So when the page number in url changes, the url changes, and the `useEffect` is triggered.
So I guess it should work... 🤔

Issue #143, opened 2021-06-07
Update task card font sizing


ankushdharkar:
Update the font sizing of the current task card to match the [figma design](https://www.figma.com/file/KAV9FsUXlzhdFbHBADiZSk/Task-Row?node-id=0%3A1)
![image](https://user-images.githubusercontent.com/1935403/120961469-bf6f1900-c712-11eb-8e39-c2de92771e02.png)

akshay1502:
I would like to this task. Assign it to me.

ankushdharkar:
@akshay1502 Assigned! Please send the PR when ready :)

PR #144, opened 2021-06-07
updated the card elements font size

akshay1502:
updated the font size of the elements according to Figma design.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 To run the project `yarn run dev`.
 
 # Production
-    
+
 To do a production build `yarn run start`.
 
 # Our websites
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 To run the project `yarn run dev`.
 
 # Production
-
+    
 To do a production build `yarn run start`.
 
 # Our websites
--- a/src/components/tasks/card/card.module.scss
+++ b/src/components/tasks/card/card.module.scss
@@ -27,19 +27,19 @@
   }
 }
 .cardTitle {
-  font-size: 24px;  
+  font-size: 1.6rem;  
   font-weight: 500;
   color: #041187;
 }
 
 .cardSpecialFont {
-  font-size: 16px;  
+  font-size: 1.1rem;  
   font-weight: 500;
   color: #aeaeae;
 }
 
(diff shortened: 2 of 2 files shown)


akshay1502:
#143

README.md
```diff
@@ -11,7 +11,7 @@
 To run the project `yarn run dev`.
 
 # Production
-
+
```
ankushdharkar, line 14: Why?

ankushdharkar (changes requested):

README.md
akshay1502: Sorry, I might have accidentally pressed enter or tab while reading the readme.md file.

ankushdharkar: Can you fix it please? And request a re-review?

akshay1502: Ok

src/components/tasks/card/card.module.scss
```diff
@@ -49,6 +49,6 @@
 }
 
 .cardStrongFont {
-  font-size: 1.5rem;
+  font-size: 1.3rem;
```
whyDontI, line 52: Can we remove extra spaces?
Also, make sure that the `rem` unit converts to whatever font-size suggested in the designs

whyDontI (changes requested):
Requesting some linting changes

src/components/tasks/card/card.module.scss
akshay1502: ok

whyDontI (changes requested):
@akshay1502 The previous changes were correct, we don't want a fixed font size for all devices.
That's why we use `em` and `rem` units

`em` is relative to the font-size of its direct or nearest parent, `rem` is only relative to the html (root) font-size.

You might want to check this blog out,
https://css-tricks.com/rem-global-em-local/

Good work 🥳 👍🏽

swarajpure:
I was comparing the previous commit to the latest commit and hence I was seeing things like 16px => 1.1rem and hence my comment

My bad.

You can keep the values according to figma design (rem only)

whyDontI (approved):
Good work, Please take care of the lint issues from next time 😃

PR #149, opened 2021-06-09
[openPRs-layout] css changes, semantic tags added

rajakvk:


rajakvk:
As per App owner, it was earlier decision to keep the page in single column. So closing PR.

PR #156, opened 2021-06-17
idlemembers: added dummy image using state

### LINKED ISSUE #151: Add a dummy picture for the members without profile picture
Use this picture
https://github.com/Real-Dev-Squad/website-status/blob/develop/public/dummyProfile.png

Screenshot 2021-06-14 at 9 09 30 PM

NoLayerCode:
Closes #151 
Hi @whyDontI ,
I have added dummy image using state. 
Please check and revert.

### DIFF
--- a/src/components/idleMembers/card/index.tsx
+++ b/src/components/idleMembers/card/index.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react';
+import { FC, useState } from 'react';
 import classNames from '@/components/idleMembers/card/card.module.scss';
 
 type Props = {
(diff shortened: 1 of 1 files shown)


src/components/idleMembers/card/index.tsx
```diff
@@ -1,12 +1,14 @@
...
+import { FC, useState } from 'react';
 import classNames from '@/components/idleMembers/card/card.module.scss';
 
 type Props = {
   idleMemberUserName: string
 }
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const imageGenerator = (name: string) => `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${name}/img.png`;
+  const [imageGenerator, setAssigneeProfilePic] = useState(`${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}${idleMemberUserName}/img.png`);
```
whyDontI, line 9: Before the above change, this state variable was a function named `imageGenerator`, now it's a variable holding assignee's profile picture.
Can we name it accordingly?

src/components/idleMembers/card/index.tsx
```diff
@@ -1,12 +1,14 @@
...
+
+  const contributorImageOnError = () => setAssigneeProfilePic('dummyProfile.png');
```
whyDontI, line 11: NIT: Can we rename it as `assigneeImageOnError`

whyDontI (changes requested):
Good work 🥳 
Requested some NIT changes

src/components/idleMembers/card/index.tsx
sumitd94: Can we please get `dummyProfile.png` from the constants file? It will be easier to maintain in the future.
even @abhinav1590 has used `dummyProfile.png`, so moving this to constants will help us. Later on, when we have to change the name of this, we can only change it in one place, rather than changing at all the places where we have hardcoded it.

whyDontI:
Solved in https://github.com/Real-Dev-Squad/website-status/pull/198

PR #157, opened 2021-06-20
Dummy picture card component

### LINKED ISSUE #152: Add a dummy picture in Card component
## Cause
https://github.com/Real-Dev-Squad/website-backend/issues/387

## Requirement
Show [dummy profile image](https://github.com/Real-Dev-Squad/website-status/blob/develop/public/dummyProfile.png) when there is no picture or a broken image URL in Card component

## Live issue
https://staging-status.realdevsquad.com/challenges

akshay1502:
Closes #152


whyDontI (changes requested):
Can you confirm that your solution won't change all the images on the page with dummy images?
If you don't find any better way to implement it, we have this solution already implemented with `stateVariables`, it'll be great if you can go through it.
https://github.com/Real-Dev-Squad/website-status/blob/292adefc59b0136ef06100aa6e9ed4e1f33873c3/src/components/tasks/card/index.tsx#L18

whyDontI (approved):
LGTM 👍🏽

PR #161, opened 2021-06-22
Removed the pointer effect of cursor

### LINKED ISSUE #159: Cursor should not be pointer if underlying component is not clickable
The cursor should be default if the card does not have a `link` in title `prop`

> You might want to wait for below GIF to load 👇🏽

nilisha-jais:
Fixes #159


whyDontI (changes requested):
Nice work 👍🏽 
Requesting some additions.
In this card component, we accept multiple props and the first one is the title
`
title: {
    text: string,
    link?: string
  }
  `
  we accept the title prop as an object where `link` is optional.
  We make the card clickable when there is a `link` provided with the title.  

  You need to make sure when there is a `link` provided make the cursor `pointer` for the whole card, and vice-versa.

nilisha-jais:
@whyDontI please check it's not working fine. Where did I went wrong

src/components/Card/index.tsx
```diff
@@ -44,7 +44,7 @@ const Card: FC<Props> = ({
 
   return (
     <div
-      className={classNames.card}
+      className={`classNames.card ${title.link ? "classNames.pointer" : ""}`}
```
whyDontI, line 47: Great, you got the gist 🥳 
But your solution won't apply class `card` because it'll be converted in the string `classNames.card`.

You might want to change it to something like below,
```
className={`${classNames.card} ${title.link && classNames.pointer}`}
```

whyDontI (changes requested):

src/components/Card/index.tsx
nilisha-jais: Got it! Thank you

correaswebert: Cool 😄 
Just one suggestion, could we use the `.links` (line 78 of scss) instead of creating a totally separate and new class?

whyDontI (approved):
Good work 👍🏽

PR #166, opened 2021-06-26
Made few enhancements on availability panel

### LINKED ISSUE #165: Few enhancements on availability panel
- [x] avoid forceful reloading of the page on task assignment 
- [x] display avatar if the image is broken
- [x] use variables in CSS
- [x] hide tasks and idle users when patch request is processing 

Design link: https://www.figma.com/file/krYzdCT45ikzDDZ2tsUQaS/Tasks-Assignment?node-id=1%3A2

Pavangbhat:
Closes #165 

- [x] avoid forceful reloading of the page on task assignment 
- [x] display avatar if the image is broken
- [x] use variables in CSS
- [x] hide tasks and idle users when patch request is processing 
- [x] added toast message instead of alert 
- [x] user will be able to sort tasks or members
- [x] only task will be able to be dropped on members and vice-versa(UX)


Pavangbhat (commented):
ghost

src/pages/availability-panel.tsx
```diff
@@ -1,55 +1,72 @@
...
+          const sortedIdleMembers = filterMembers.sort();
+          setIdleMembersList(sortedIdleMembers);
+        }
+        setIsLoading(false);
+      } catch (Error) {
+        setError(true);
+        setIsLoading(false);
+      }
+    };
+    fetchTasks();
```
Pavangbhat, line 51: Should I make this IIFE?

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -42,27 +42,27 @@ const DragDropcontext: FC<dragDropProps> = ({
           ? 'Sucessfully Assigned Task'
           : 'Something went wrong';
```
whyDontI, line 43: This condition will never be false because any error returned from API will take the execution in the catch block.
Do the needful

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -42,27 +42,27 @@ const DragDropcontext: FC<dragDropProps> = ({
...
         <div className={classNames.statusMessage}>Please wait...</div>
       )}
+      {!isProcessing && (
       <div className={classNames.flexContainer}>
         <div>
           {unAssignedTasks.length === 0 ? (
-            <div className={classNames.emptyArray}>No Tasks found</div>
+            <div className={classNames.emptyArray}>
+              <img src="ghost.png" alt="ghost" />
+              No Tasks found
```
whyDontI, line 64: Style this message as shown in the designs

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -89,7 +89,10 @@ const DragDropcontext: FC<dragDropProps> = ({
         <div className={classNames.divider} />
         <div>
           {idleMembers.length === 0 ? (
-            <div className={classNames.emptyArray}>No idle users Found</div>
+            <div className={classNames.emptyArray}>
+              <img src="ghost.png" alt="ghost" />
+              No idle users Found
```
whyDontI, line 94: Style the text as designs

whyDontI (changes requested):

src/pages/availability-panel.tsx
whyDontI: No, it's good.
Just move the cleanup like setting loading false and clearing the errors in `finally` block

whyDontI:
Also, can you put a separator for mobile view?

whyDontI (changes requested):
Good work, requesting some changes
- I think you need to reset the `isProcessing` state variable after error
- Also can you show the error message coming from the API response in the toast?

Jul-15-2021 23-00-05

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -81,7 +81,7 @@ const DragDropcontext: FC<dragDropProps> = ({
           ? result.draggableId
           : result.combine.draggableId;
         const data = {
-          status: ASSIGNED,
+          status: ACTIVE,
```
Pavangbhat, line 84: Status must be set to 'active' else the assigned member will not be removed from the idle members list

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -90,8 +90,7 @@ const DragDropcontext: FC<dragDropProps> = ({
           method: 'patch',
           data,
         });
-
-        toast.success(`${updateTaskResult.data.message}`, {
+        toast.success(updateTaskResult.data.message ? updateTaskResult.data.message : 'Sucessfully Assigned Task', {
```
Pavangbhat, line 93: We don't have any message from the response when its PATCH request is successful

src/components/availability-panel/drag-drop-context/index.tsx
whyDontI: Then we'll have to make a new issue in the backend to change the logic for IdleMembers.
Because a user can have multiple tasks ASSIGNED to them but they'll only work actively on one task.
Which will have the status ACTIVE.

Refer [this](https://discord.com/channels/673083527624916993/735063616168001536/848892753004986369) discussion

Pavangbhat: Right, But for the availability panel, we need to do something to handle both

src/components/availability-panel/drag-drop-context/index.tsx
whyDontI: Let's remove the `updateTaskResult.data.message` and just keep hard coded message.
Just checked the API contracts `204` doesn't return a response 👍🏽 
https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/tasks#patch-tasksid

whyDontI (approved):

PR #169, opened 2021-06-29
cancel pending requests if user navigates away from the page

### LINKED ISSUE #51: Cancel pending requeset if user navigates away from page
## Problem
When the user navigates away from the component when a request is in the pending state, we no longer care about the result and want to cancel a request that’s already sent

<a href="https://www.loom.com/share/4355d84b5ade44e4b18ef161aa6f3a64" target="_blank">Click to watch video of the problem</a>

### Docs
<a href="https://www.npmjs.com/package/axios#cancellation" target="_blank">Axios Cancellation docs</a>

Ishikabaid:
closes #51

### DIFF
--- a/src/hooks/useFetch.ts
+++ b/src/hooks/useFetch.ts
@@ -5,25 +5,32 @@ const useFetch = (url: string, options: object = {}) => {
   const [response, setResponse] = useState<any>({});
   const [error, setError] = useState<any>(null);
   const [isLoading, setIsLoading] = useState(true);
-
+  let cancel: () => void;
   useEffect(() => {
     (async () => {
       setIsLoading(true);
       try {
-        const res = await fetch({
+        const { response: res, cancelApi } = fetch({
           url,
           method: 'get',
           ...options,
         });
-        setResponse(res.data);
+        cancel = cancelApi;
+        const fetchPromise = await res;
+        setResponse(fetchPromise.data);
       } catch (err) {
         setError(err);
       } finally {
         setIsLoading(false);
       }
     })();
+    return () => {
+      cancel();
+    };
   }, [url]);
-  return { response, error, isLoading };
+  return {
+    response, error, isLoading,
+  };
 };
 
 export default useFetch;
--- a/src/hooks/useFetch.ts
+++ b/src/hooks/useFetch.ts
@@ -10,7 +10,7 @@ const useFetch = (url: string, options: object = {}) => {
     (async () => {
       setIsLoading(true);
       try {
-        const { response: res, cancelApi } = fetch({
+        const { requestPromise: res, cancelApi } = fetch({
           url,
           method: 'get',
           ...options,
--- a/src/hooks/useFetch.ts
+++ b/src/hooks/useFetch.ts
@@ -10,13 +10,13 @@ const useFetch = (url: string, options: object = {}) => {
     (async () => {
       setIsLoading(true);
       try {
-        const { requestPromise: res, cancelApi } = fetch({
+        const { requestPromise, cancelApi } = fetch({
           url,
           method: 'get',
           ...options,
         });
         cancel = cancelApi;
-        const fetchPromise = await res;
+        const fetchPromise = await requestPromise;
         setResponse(fetchPromise.data);
       } catch (err) {
         setError(err);
--- a/src/pages/availability-panel.tsx
+++ b/src/pages/availability-panel.tsx
@@ -18,8 +18,9 @@ const AvailabilityPanel: FC = () => {
     const fetchTasks = async () => {
       try {
         const url = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
-        const response = await fetch({ url });
-        const { tasks } = response.data;
+        const { requestPromise } = fetch({ url });
+        const fetchPromise = await requestPromise;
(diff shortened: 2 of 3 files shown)


whyDontI (changes requested):
The task is to cancel requests when the user navigates away from the page, which means we should be canceling requests when the component unmounts.

Plus we use [this function](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/helperFunctions/fetch.ts) for making requests, you need to return a cancel token from this function.
Then we can cancel the request in the component when it unmounts.
Let me know if you need any help with this

src/components/PullRequestList/index.tsx
```diff
@@ -73,6 +73,16 @@ const PullRequestList: FC<PullRequestListProps> = ({ prType }) => {
     );
   });
 
+  const { CancelToken } = axios;
+  let cancel;
+
+  axios.get('https://api.realdevsquad.com/user', {
+    cancelToken: new CancelToken((c) => {
+      cancel = c;
```
correaswebert, line 81: You cannot assign a function scoped variable inside a callback scope.
The reason being, the callback is offloaded to a worker thread and the main thread progresses.
So it is essentially a race condition, whether the worker finishes before the main thread arrives at the utilization of the assigned part.
In this case it is guaranteed to fail, because of such less time to reach the utilization.
Learning the hard way will cement this in your memory like me 😂

whyDontI:
Can you please solve the merge conflicts

src/components/PullRequestList/index.tsx
```diff
@@ -26,10 +26,13 @@ const PullRequestList: FC<PullRequestListProps> = ({ prType }) => {
...
   const prUrl = `${process.env.NEXT_PUBLIC_BASE_URL}/pullrequests/${prType}?page=${page}`;
 
+  const { CancelToken } = axios;
+  const source = CancelToken.source();
+
   const fetchPRs = async () => {
     try {
       setLoading(true);
-      const response = await axios.get(prUrl);
+      const response = await axios({ url: prUrl, cancelToken: source.token });
```
whyDontI, line 35: We don't use `axios` or any request library in application code.

We have a [helper function](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/helperFunctions/fetch.ts) for the same.

We also have a [custom hook](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/hooks/useFetch.ts) to make requests.

You'll have to write this cancellation logic in both of them.

whyDontI (changes requested):

src/components/PullRequestList/index.tsx
Ishikabaid: Yes made all the necessary changes.

src/components/PullRequestList/index.tsx
Ishikabaid: Yes, it was an outdated code.
 All the necessary changes have been made now. 
Thanks for your suggestions!

src/pages/availability-panel.tsx
```diff
@@ -18,8 +18,9 @@ const AvailabilityPanel: FC = () => {
     const fetchTasks = async () => {
       try {
         const url = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
-        const response = await fetch({ url });
-        const { tasks } = response.data;
+        const { requestPromise } = fetch({ url });
+        const fetchPromise = await requestPromise;
+        const { tasks } = fetchPromise.data;
```
swarajpure, line 23: Any specific reason to not use the `useFetch` hook? It is anyway having fetching + cancelling functionality, right?

src/hooks/useFetch.ts
```diff
@@ -5,25 +5,32 @@ const useFetch = (url: string, options: object = {}) => {
...
+        const fetchPromise = await res;
+        setResponse(fetchPromise.data);
       } catch (err) {
         setError(err);
       } finally {
         setIsLoading(false);
       }
     })();
+    return () => {
+      cancel();
```
swarajpure, line 28: Sweet!

swarajpure (changes requested):

src/hooks/useFetch.ts
```diff
@@ -5,25 +5,32 @@ const useFetch = (url: string, options: object = {}) => {
...
   const [error, setError] = useState<any>(null);
   const [isLoading, setIsLoading] = useState(true);
-
+  let cancel: () => void;
   useEffect(() => {
     (async () => {
       setIsLoading(true);
       try {
-        const res = await fetch({
+        const { requestPromise: res, cancelApi } = fetch({
```
whyDontI, line 13: Any reason for renaming the `requestPromise`?

src/pages/availability-panel.tsx
whyDontI: Yeah, `useFetch` can be used.
But that change is not related to this issue or pr, let's create a new refactoring issue for this.

src/hooks/useFetch.ts
Ishikabaid: Oops yeah I should not be doing this now.
I'll remove res and use requestPromise only.

whyDontI (approved):
Great work @Ishikabaid 🥳

PR #179, opened 2021-08-12
Updated PR loading on bigger screens - issue

### LINKED ISSUE #174: PRs doesn't load on bigger screens
On [OpenPR](https://staging-status.realdevsquad.com/openPRs) and [StalePR](https://staging-status.realdevsquad.com/stale-pr) pages we have infinite scroll implemented.
As the scroller hits the bottom page loads the next set of PRs.
This doesn't seem to work on bigger screens.

Screenshot 2021-08-03 at 11 07 48 AM

rohandhar1234:
Closes #174 

Works fine for 
1920 x 1080 (FHD)
3840 x 2160 (4K)
7680 x 4320 (8K)

### DIFF
--- a/src/components/PullRequestList/index.tsx
+++ b/src/components/PullRequestList/index.tsx
@@ -37,7 +37,7 @@ const PullRequestList: FC<PullRequestListProps> = ({ prType }) => {
   const [page, setPage] = useState(1);
   const [isBottom, setIsBottom] = useState(false);
 
-  const prUrl = `${process.env.NEXT_PUBLIC_BASE_URL}/pullrequests/${prType}?page=${page}`;
+  const prUrl = `${process.env.NEXT_PUBLIC_BASE_URL}/pullrequests/${prType}?page=${page}&size=100`;
   const {
     response,
     error,
--- a/src/components/PullRequestList/index.tsx
+++ b/src/components/PullRequestList/index.tsx
@@ -36,8 +36,9 @@ const PullRequestList: FC<PullRequestListProps> = ({ prType }) => {
   const [noData, setNoData] = useState(false);
   const [page, setPage] = useState(1);
   const [isBottom, setIsBottom] = useState(false);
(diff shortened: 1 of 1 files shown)


rohandhar1234:
1920 x 1080

fhd
4K
4k
8K

8k

rohandhar1234:
I set the elements that each page displays to 100. This way be it any display size, the PR will always load and the scroll mechanism will work with it. This also works with 65 elements being displayed on each page but I still set it to 100 because this way we can cover a lot of data and can view it on displays greater than 8K

whyDontI (changes requested):
Technically this solution works 😃 👍🏽 

But, we'll load all 100 PRs even on the small screens.

Can you calculate the number of cards needed using the screen height and card height to fill the screen and request that many cards only?

The formula might look like this,
```
const numberOfCards = ( screenHeight / cardHeight ) * 3
// 3 - is the number of cards in a row 
```

src/components/PullRequestList/index.tsx
```diff
@@ -36,8 +36,9 @@ const PullRequestList: FC<PullRequestListProps> = ({ prType }) => {
   const [noData, setNoData] = useState(false);
   const [page, setPage] = useState(1);
   const [isBottom, setIsBottom] = useState(false);
+  const size = Math.floor((global.innerHeight / 147) * 3);
```
whyDontI, line 39: Normally `global` is for the node's runtime and `window` is for browsers. I am not sure how it's working here, I'll try to figure that out.
But for now let's change `global` to `globalThis` so there is no confusion of this sort.

> Recently, globalThis was added to the language, as a standardized name for a global object, that should be supported across all environments. It’s supported in all major browsers.
Refer - https://javascript.info/global-object

whyDontI (changes requested):

rohandhar1234:
Closes #174

PR #184, opened 2021-08-19
Excluded 'UNASSIGNED' tasks from overdue highlight

### LINKED ISSUE #173: Exclude "UNASSIGNED" tasks from overdue highlight
Currently, all tasks (except "COMPLETED") which have overshot their deadlines are being given a red border and highlighted. We want the "UNASSIGNED" tasks to be excluded from the overdue tasks list (similar to "COMPLETED" tasks).

A picture of current "UNASSIGNED" tasks - 
image

Expected picture of "UNASSIGNED" tasks after this issue is fixed - 
image

abhinav1590:
Closes #173

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,3 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,6 +1,3 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
-
-// NOTE: This file should not be edited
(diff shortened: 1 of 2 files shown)


next-env.d.ts
```diff
@@ -1,3 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/types/global" />
 /// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
```
harshith-venkatesh, line 5: What is purpose of this comment here.

harshith-venkatesh (changes requested):

next-env.d.ts
abhinav1590: I have no idea how it got there..

whyDontI: Don't delete this file, it was already there. Just don't make the changes

whyDontI: That's a Next js generated file.
It might get updated on `yarn install`.

Run `yarn install` and push the updated version of the file.

Also, you should check what files you are adding to your commits.

> I have no idea how it got there..

☝🏽This is not a good answer

PR #185, opened 2021-08-22
Fixing case sensitive capitalization in sorting idle members

### LINKED ISSUE #167: Arrange the cards in idle members section alphabetically
Currently, we see the members are sorted alphabetically but there is a error here due to which we see two names at top.
The effort involves reading about ascii codes, reading sort method in JS and adding relevant changes

https://media.discordapp.net/attachments/741616265050849282/859129265554587718/unknown.png?width=471&height=473

rohan09-raj:
Closes #167


whyDontI (approved):
Thanks, @rohan09-raj 🥳

PR #186, opened 2021-08-23
Excluded 'UNASSIGNED' tasks from overdue highlight

### LINKED ISSUE #173: Exclude "UNASSIGNED" tasks from overdue highlight

abhinav1590:
closes #173

### DIFF
--- a/src/components/tasks/card/index.tsx
+++ b/src/components/tasks/card/index.tsx
@@ -6,6 +6,7 @@ import {
   ACTIVE,
   ASSIGNED,
   COMPLETED,
+  UNASSIGNED,
 } from '@/components/constants/task-status';
 
 const moment = require('moment');
@@ -38,7 +39,7 @@ const Card: FC<Props> = ({ content }) => {
   function isTaskOverdue() {
     const currentDate = new Date();
     const timeLeft = localEndsOn.valueOf() - currentDate.valueOf();
-    return status !== COMPLETED && timeLeft <= 0;
+    return status !== COMPLETED && status !== UNASSIGNED && timeLeft <= 0;
   }
 
   if (isTaskOverdue()) {


src/components/tasks/card/index.tsx
```diff
@@ -38,7 +39,7 @@ const Card: FC<Props> = ({ content }) => {
   function isTaskOverdue() {
     const currentDate = new Date();
     const timeLeft = localEndsOn.valueOf() - currentDate.valueOf();
-    return status !== COMPLETED && timeLeft <= 0;
+    return status !== COMPLETED && status !== UNASSIGNED && timeLeft <= 0;
```
whyDontI, line 42: This should be `OR / ||` right?
Status can't be COMPLETED and UNASSIGNED at the same time.

whyDontI (changes requested):

src/components/tasks/card/index.tsx
abhinav1590: image
This is the result when I am using `AND / &&` operator 

image
And this ,when I use `OR / || `operator

whyDontI: The function should return `true` if the task is overdue and `false` otherwise.
Check if the function works properly and try to fix the implementation if there is any problem with the same.

Doesn't matter even if the output is correct when we can't understand the implementation.

whyDontI (approved):
Good work @abhinav1590 👍🏽 🥳

PR #193, opened 2021-08-31
Challenges button integrated with api

akshay1502:


src/components/challenges/active.tsx
```diff
@@ -2,6 +2,9 @@ import { FC, useState } from 'react';
 import Card from '@/components/Card/index';
 import details from '@/components/challenges/details';
 import participantsDetails from '@/components/challenges/participants';
+import { toast } from 'react-toastify';
+
+const SUBSCRIBE_TO_CHALLENGE_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/challenges/subscribe`;
```
sumitd94, line 7: I think it will be good if we move this to an AppConstants file that holds all our application constants.
cc - @whyDontI

src/components/challenges/active.tsx
```diff
@@ -27,26 +30,49 @@ type ActiveProps = {
...
+      .catch((err) => {
+        toast.error(err, {
+          position: 'top-center',
+          autoClose: 4000,
+          hideProgressBar: false,
+          closeOnClick: true,
+          pauseOnHover: false,
+          draggable: false,
+          progress: undefined,
+        });
```
sumitd94, line 61: This needs to be placed outside of this. what if we want to use the toast elsewhere? are we going to write this again?

Instead create a function that calls this, and invoke this function with the error message.

src/pages/challenges.tsx
```diff
@@ -9,10 +9,11 @@ import challenge from '@/interfaces/challenge.type';
 import classNames from '@/styles/tasks.module.scss';
 
 const CHALLENGES_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/challenges`;
+const USER_SELF = `${process.env.NEXT_PUBLIC_BASE_URL}/users/self`;
```
sumitd94, line 12: move to constants file, please

src/components/challenges/active.tsx
whyDontI: Yeah, we need to get someone to build these toast functions.
I'll create an issue 👍🏽 
Thanks

src/components/challenges/active.tsx
whyDontI: @akshay1502 Please create an `urls.js` file in the `components/constants` directory and move these variables there. 👍🏽

whyDontI:
Toast service has been created, I think we should resume working on this.

src/components/challenges/active.tsx
```diff
@@ -27,26 +29,43 @@ type ActiveProps = {
...
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(data),
+      credentials: 'include',
+    })
+      .then((res) => res.json())
+      .then((res) => setUserSubscribed(res.is_user_subscribed))
+      .catch((err) => {
+        toast(ToastTypes.ERROR, err);
+      });
+  };
```
whyDontI, line 56: We are using async-await in the codebase everywhere, let's make this the same.

whyDontI: Also, put the code in `try-catch` block, show the toast on error or success.

whyDontI (changes requested):
Requesting some small changes

src/components/challenges/active.tsx
```diff
@@ -27,26 +29,46 @@ type ActiveProps = {
...
+  const subscibeUser = async () => {
+    const data = {
+      challenge_id: content.id,
+      user_id: userId,
+    };
+    try {
+      const response = await fetch(SUBSCRIBE_TO_CHALLENGE_URL, {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(data),
```
whyDontI, line 50: Why are we stringifying data?
Does API not work with JSON?

src/components/challenges/active.tsx
```diff
@@ -27,26 +29,46 @@ type ActiveProps = {
...
+const Active: FC<ActiveProps> = ({ content, userId }) => {
   const [isUserSubscribed, setUserSubscribed] = useState(content.is_user_subscribed);
 
+  const subscibeUser = async () => {
+    const data = {
+      challenge_id: content.id,
+      user_id: userId,
+    };
+    try {
+      const response = await fetch(SUBSCRIBE_TO_CHALLENGE_URL, {
```
whyDontI, line 47: We have our own implementation of the `fetch` [here](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/helperFunctions/fetch.ts), please use the same.

src/components/challenges/active.tsx
```diff
@@ -27,26 +29,46 @@ type ActiveProps = {
...
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify(data),
+        credentials: 'include',
+      });
+      const { isUserSubscribed: subscription } = await response.json();
```
whyDontI, line 53: We emit this step after using our own `fetch`, as requested in above comment.

src/components/challenges/active.tsx
```diff
@@ -27,26 +29,46 @@ type ActiveProps = {
...
+      setUserSubscribed(subscription);
+      toast(SUCCESS, 'You have subscribed to the challenge');
+    } catch (error:any) {
+      toast(ERROR, error.message);
```
whyDontI, line 57: If an error has `response` present, please toast the `error.response.message` which is the message returned from the RDS backend.

src/components/constants/url.js
```diff
@@ -0,0 +1,4 @@
+export const SUBSCRIBE_TO_CHALLENGE_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/challenges/subscribe`;
+
```
whyDontI, line 2: Remove this blank line, if not intentional

src/pages/challenges.tsx
```diff
@@ -23,7 +22,7 @@ const Challenges: FC = () => {
   const [filteredChallenge, setFilteredChallenge] = useState<any>([]);
 
   const { response, error, isLoading } = useFetch(CHALLENGES_URL);
-
+  const { response: user } = useFetch(USER_SELF);
```
whyDontI, line 25: Two responses?
And we show all the challenges on this page, there is another page for personal challenges.

whyDontI (changes requested):
Apologies for the late review.
Requesting some changes.

src/pages/challenges.tsx
akshay1502: when the user subscribes to the challenges we are passing the userId as data. So as to get userId I am calling users/self/

whyDontI: The second `useFetch` call will overwrite the `response` constant of the first `useFetch` call.
A good approach will be, writing a separate function to get the details of the logged-in user, and use that in a `useEffect` to get the details at the component mounting phase.

PR #196, opened 2021-09-03
seperate service for toast

akshay1502:


src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,13 @@
+import { toast } from 'react-toastify';
+
+type ToastType = 'info' | 'success' | 'warning' | 'error' ;
```
sumitd94, line 3: Can you please add an empty line after this?

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,13 @@
...
+const toastContainer = (type: ToastType, message: string) => toast[type](message, {
+  position: toast.POSITION.TOP_RIGHT,
+  autoClose: 3000,
+  hideProgressBar: false,
+  closeOnClick: true,
+  pauseOnHover: false,
+  draggable: false,
+  progress: undefined,
+});
```
sumitd94, line 12: Re:

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,13 @@
+import { toast } from 'react-toastify';
+
+type ToastType = 'info' | 'success' | 'warning' | 'error' ;
+const toastContainer = (type: ToastType, message: string) => toast[type](message, {
+  position: toast.POSITION.TOP_RIGHT,
+  autoClose: 3000,
+  hideProgressBar: false,
```
Pavangbhat, line 7: Please check what options are there by default with **react-toastify** and probably we can remove them. For example, if **hideProgressBar** is **false** by default we can avoid writing it explicitly

Pavangbhat, line 7: Please check what options are there by default with **react-toastify** and probably we can remove them. For example, if **hideProgressBar** is **false** by default we can avoid writing it explicitly

Pavangbhat (commented):
Please check what options are there by default with **react-toastify** and probably we can remove them. For example, if **hideProgressBar** is **false** by default we can avoid writing it explicitly

sumitd94 (approved):

Pavangbhat (dismissed):

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,12 @@
+import { toast } from 'react-toastify';
+
+type ToastType = 'info' | 'success' | 'warning' | 'error' ;
+
```
whyDontI, line 4: Can we please add [TypeScript enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums)?
Also, export it

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,12 @@
...
+const Toast = (type: ToastType, message: string) => toast[type](message, {
```
whyDontI, line 5: Any reason for the capital first letter? 🤔 
If not please make it lowercase, we follow the `camelCase` variable name in this codebase.

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,12 @@
...
+  position: toast.POSITION.TOP_RIGHT,
+  autoClose: 3000,
```
whyDontI, line 7: Make it a variable at the start of the file

whyDontI (changes requested):
Good work @akshay1502 👍🏽 
Requesting some changes.

Also, can you please use this service wherever we have used toast in the code?

src/helperFunctions/toast.ts
akshay1502: naming it as **toast** gives an error as _import declaration conflicts with local declaration of 'toast'._

.eslintrc.json
```diff
@@ -53,7 +53,11 @@
               "**/*.stories.tsx"
             ]
           }
-        ]
+        ],
+        "no-shadow": "off",
+        "@typescript-eslint/no-shadow": ["error"],
```
akshay1502, line 58: it's a TSlint to ESlint bug and to fix it these configuration was added.
[StackOverflow link](https://stackoverflow.com/a/65768375/13964048)

.eslintrc.json
```diff
@@ -53,7 +53,11 @@
...
+        "no-unused-vars": "off",
+        "@typescript-eslint/no-unused-vars": "error"
```
akshay1502, line 60: i was getting enum declared but not used error even after using it.
Found this configuration as solution to it.
https://github.com/typescript-eslint/typescript-eslint/issues/2621#issuecomment-701970389

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,21 @@
+import { toast } from 'react-toastify';
```
whyDontI, line 1: Let's make this as `import { toast as toastify } from ....` and rename our function name to `toast`

src/helperFunctions/toast.ts
whyDontI: Addressed the issue in the comment on line number 1

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const DELAY: number = 3000;
+enum ToastTypes {
+  INFO = 'info',
+  SUCCESS = 'success',
+  WARNING = 'warning',
+  ERROR = 'error'
+}
+
+type ToastType = ToastTypes ;
```
whyDontI, line 11: No need to do this, we can directly use it.

src/helperFunctions/toast.ts
```diff
@@ -0,0 +1,21 @@
+import { toast } from 'react-toastify';
+
+const DELAY: number = 3000;
+enum ToastTypes {
```
whyDontI, line 4: Can we export from here?

whyDontI (changes requested):
Requesting some changes

Pavangbhat (changes requested):
Can we destructure ToastTypes ? If yes please make it

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -2,7 +2,7 @@ import { FC, useState, useEffect } from 'react';
 import classNames from '@/components/availability-panel/drag-drop-context/styles.module.scss';
 import { DragDropContext, DropResult } from 'react-beautiful-dnd';
 import { dragDropProps } from '@/interfaces/availabilityPanel.type';
-import { toast } from 'react-toastify';
+import toast, { ToastTypes } from '@/helperFunctions/toast';
```
Pavangbhat, line 5: Probably good to avoid default export here and export it as const isn't it @sumitd94 @whyDontI ?

Pavangbhat (changes requested):

sumitd94 (dismissed):

Pavangbhat (approved):
Looks good to me.
Great work 🎉🎉

sumitd94 (approved):

PR #198, opened 2021-09-09
Dummy image added

### LINKED ISSUE #151: Add a dummy picture for the members without profile picture

abhinav1590:
Closes #151
Addressed the comments and made the requested changes

### DIFF
--- a/src/components/constants/display-sections.js
+++ b/src/components/constants/display-sections.js
@@ -1,3 +1,2 @@
 export const CHALLENGES_SCREEN = 'Challenges';
 export const DUMMY_PROFILE = 'dummyProfile.png';
-export const IMAGE_URL = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
--- a/src/components/constants/display-sections.js
+++ b/src/components/constants/display-sections.js
@@ -1,1 +1,2 @@
 export const CHALLENGES_SCREEN = 'Challenges';
+export const dummyProfile = 'dummyProfile.png';
--- a/src/components/constants/display-sections.js
+++ b/src/components/constants/display-sections.js
@@ -1,2 +1,3 @@
 export const CHALLENGES_SCREEN = 'Challenges';
 export const DUMMY_PROFILE = 'dummyProfile.png';
+export const IMAGE_URL = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
--- a/src/components/constants/display-sections.js
+++ b/src/components/constants/display-sections.js
@@ -1,2 +1,2 @@
 export const CHALLENGES_SCREEN = 'Challenges';
-export const dummyProfile = 'dummyProfile.png';
+export const DUMMY_PROFILE = 'dummyProfile.png';
--- a/src/components/idleMembers/card/index.tsx
+++ b/src/components/idleMembers/card/index.tsx
@@ -14,7 +14,7 @@ const Card: FC<Props> = ({ idleMemberUserName }) => {
     const newWindow = window.open(`https://members.realdevsquad.com/${name}`, '_blank', ' noopener ,norefferrer');
     if (newWindow) newWindow.opener = null;
   };
-  const assigneeImageOnError = (e) => {
+  const assigneeImageOnError = (e: any) => {
     e.target.src = dummyProfile;
   };
   return (
--- a/src/components/idleMembers/card/index.tsx
+++ b/src/components/idleMembers/card/index.tsx
@@ -13,7 +13,45 @@ const Card: FC<Props> = ({ idleMemberUserName }) => {
   const getMemberDetails = (name: string) => {
     const newWindow = window.open(`https://members.realdevsquad.com/${name}`, '_blank', ' noopener ,norefferrer');
     if (newWindow) newWindow.opener = null;
+  };import { FC, SyntheticEvent } from 'react';
+import classNames from '@/components/idleMembers/card/card.module.scss';
+import { dummyProfile } from '@/components/constants/display-sections.js';
+
+const imgUrl = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
+
+type Props = {
+  idleMemberUserName: string
+}
+
+const Card: FC<Props> = ({ idleMemberUserName }) => {
+  const assigneeProfilePic = (name: string) => `${imgUrl}/${name}/img.png`;
+  const getMemberDetails = (name: string) => {
+    const newWindow = window.open(`https://members.realdevsquad.com/${name}`, '_blank', ' noopener ,norefferrer');
+    if (newWindow) newWindow.opener = null;
   };
+  const assigneeImageOnError = (e: SyntheticEvent<HTMLImageElement>) => {
+    e.currentTarget.src = dummyProfile;
+  };
+  return (
+    <div
+      className={classNames.card}
+      onClick={() => getMemberDetails(idleMemberUserName)}
+      onKeyDown={() => getMemberDetails(idleMemberUserName)}
+      aria-hidden="true"
+    >
+      <img
+        className={classNames.image}
+        src={assigneeProfilePic(idleMemberUserName)}
+        alt={idleMemberUserName}
+        onError={assigneeImageOnError}
+      />
+      <span className={classNames.name}>{idleMemberUserName}</span>
+    </div>
+  );
+};
+
+export default Card;
+
   const assigneeImageOnError = (events: SyntheticEvent<HTMLImageElement>) => {
     events.currentTarget.src = dummyProfile;
   };
--- a/src/components/idleMembers/card/index.tsx
+++ b/src/components/idleMembers/card/index.tsx
@@ -1,21 +1,21 @@
-import { FC } from 'react';
+import { FC, SyntheticEvent } from 'react';
 import classNames from '@/components/idleMembers/card/card.module.scss';
+import { dummyProfile } from '@/components/constants/display-sections.js';
 
-const dummyProfile = 'dummyProfile.png';
-const url = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
+const imgUrl = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
 
 type Props = {
   idleMemberUserName: string
 }
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const assigneeProfilePic = (name: string) => `${url}/${name}/img.png`;
+  const assigneeProfilePic = (name: string) => `${imgUrl}/${name}/img.png`;
   const getMemberDetails = (name: string) => {
     const newWindow = window.open(`https://members.realdevsquad.com/${name}`, '_blank', ' noopener ,norefferrer');
     if (newWindow) newWindow.opener = null;
   };
-  const assigneeImageOnError = (e: any) => {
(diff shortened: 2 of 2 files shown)


rohandhar1234 (changes requested):
The source of the images is not being accessed correctly. Members with an active image are also being displayed with a dummy.png photo.

src/components/idleMembers/card/index.tsx
```diff
@@ -1,13 +1,13 @@
...
 import classNames from '@/components/idleMembers/card/card.module.scss';
 
 type Props = {
   idleMemberUserName: string
 }
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const imageGenerator = (name: string) => `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${name}/img.png`;
-
+  const [assigneeProfilePic, setAssigneeProfilePic] = useState(`${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}${idleMemberUserName}/img.png`);
```
rohandhar1234, line 9: Instead of **https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members/ankita/img.png**, your code shows **https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/membersankita/img.png**
You probably missed a **'/'** here.
It should be changed to:
const [assigneeProfilePic, setAssigneeProfilePic] = useState(`${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${idleMemberUserName}/img.png`);

abhinav1590: Ohh yeah 😅. Got it

rohandhar1234 (approved):

src/components/idleMembers/card/index.tsx
```diff
@@ -1,13 +1,13 @@
...
 
 type Props = {
   idleMemberUserName: string
 }
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const imageGenerator = (name: string) => `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${name}/img.png`;
-
+  const [assigneeProfilePic, setAssigneeProfilePic] = useState(`${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${idleMemberUserName}/img.png`);
+  const assigneeImageOnError = () => setAssigneeProfilePic('dummyProfile.png');
```
akshay1502, line 10: Can you add dummyProfile image in a separate variable? And use that variable over here.

akshay1502: Also, make another variable for URL inside of useState.

src/components/idleMembers/card/index.tsx
```diff
@@ -1,13 +1,13 @@
...
 import classNames from '@/components/idleMembers/card/card.module.scss';
 
 type Props = {
   idleMemberUserName: string
 }
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const imageGenerator = (name: string) => `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${name}/img.png`;
-
+  const [assigneeProfilePic, setAssigneeProfilePic] = useState(`${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${idleMemberUserName}/img.png`);
```
Pavangbhat, line 9: Any specific reason to use state management here ? 
check this out :point_down: 
**https://github.com/Real-Dev-Squad/website-members/blob/develop/src/components/member-card/index.js**

Pavangbhat (commented):
We can avoid using state management here

src/components/idleMembers/card/index.tsx
abhinav1590: I am not sure why the original editor used state management😅, my task was only to make small changes in this PR  https://github.com/Real-Dev-Squad/website-status/pull/156

Pavangbhat: Got it. @whyDontI can't we avoid using state management here ?

akshay1502: Yes. We can directly have idlememberURL or if an error then we can have dummyProfileURL. I think we can avoid useState.

src/components/idleMembers/card/index.tsx
```diff
@@ -1,17 +1,22 @@
 import { FC } from 'react';
 import classNames from '@/components/idleMembers/card/card.module.scss';
 
+const dummyProfile = 'dummyProfile.png';
```
sumitd94, line 4: move this to the constants file, please.

src/components/idleMembers/card/index.tsx
```diff
@@ -1,17 +1,22 @@
...
+const url = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
```
sumitd94, line 5:
```suggestion
const imgUrl = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
```
`url` is very vague here, we can rename it to something more meaningful. What say?

src/components/idleMembers/card/index.tsx
```diff
@@ -1,17 +1,22 @@
...
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const imageGenerator = (name: string) => `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${name}/img.png`;
-
+  const assigneeProfilePic = (name: string) => `${url}/${name}/img.png`;
   const getMemberDetails = (name: string) => {
     const newWindow = window.open(`https://members.realdevsquad.com/${name}`, '_blank', ' noopener ,norefferrer');
     if (newWindow) newWindow.opener = null;
   };
+  const assigneeImageOnError = (e: any) => {
```
sumitd94, line 17: why `any`? Please avoid using `any`, it totally defeats the purpose of using typescript.

Here `e` can be - `SyntheticEvent<HTMLDivElement>` (Just re-check once).
Also, replace `e` with `events`.

src/components/constants/display-sections.js
```diff
@@ -1 +1,2 @@
 export const CHALLENGES_SCREEN = 'Challenges';
+export const dummyProfile = 'dummyProfile.png';
```
Pavangbhat, line 2: Please make **dummyProfile** as upper case and words must be separated by underscores

Pavangbhat (changes requested):
Please make **dummyProfile** uppercase. Rest everything looks good to me.
Nice work :tada:

src/components/constants/display-sections.js
swarajpure: The idea here is that it's a convention (not a rule) that we keep the those variables in all uppercase -- whose value is known to us before the program starts executing, is not calculated during compile time and does not change throughout the execution. It's easier for a developer to understand the code while going through it "okay, this is a constant known to us beforehand and is not calculated during runtime, its value must be declared somewhere, let me go and check"

Hope that makes sense @abhinav1590 

Also @Pavangbhat good that you pointed it out but please mention reasons also, as to why you think something should be done. You don't have to type huge paragraphs but a one-liner or a blog/article/github or stackoverflow thread would also do the job. The point is that everyone should understand why they're doing what they're doing, not just because some reviewer told them to do so.

Pavangbhat: Thank you @swarajpure , Yes I should have explained it. Next time for sure will keep this in mind

src/components/idleMembers/card/index.tsx
```diff
@@ -1,17 +1,22 @@
-import { FC } from 'react';
+import { FC, SyntheticEvent } from 'react';
 import classNames from '@/components/idleMembers/card/card.module.scss';
+import { DUMMY_PROFILE } from '@/components/constants/display-sections.js';
+
+const imgUrl = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
```
swarajpure, line 5: Now that you know the convention for constant values being uppercase, do you feel it should be used here as well? 😉

src/components/idleMembers/card/index.tsx
```diff
@@ -1,17 +1,22 @@
...
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const imageGenerator = (name: string) => `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${name}/img.png`;
-
+  const assigneeProfilePic = (name: string) => `${imgUrl}/${name}/img.png`;
   const getMemberDetails = (name: string) => {
     const newWindow = window.open(`https://members.realdevsquad.com/${name}`, '_blank', ' noopener ,norefferrer');
     if (newWindow) newWindow.opener = null;
   };
+  const assigneeImageOnError = (e: SyntheticEvent<HTMLImageElement>) => {
+    e.currentTarget.src = DUMMY_PROFILE;
```
swarajpure, line 18: Are you sure this works? Can you please provide a screenshot of the frontend where this dummy image is being used?

src/components/idleMembers/card/index.tsx
abhinav1590: I have to put it in constants folder and uppercase imgUrl , I guess😅

src/components/idleMembers/card/index.tsx
abhinav1590: image
Here it is, when I purposely create an error this is the frontend I get

abhinav1590: And if I use `e.target.src = DUMMY_PROFILE;` then I am getting an error - Property 'src' does not exist on type 'EventTarget'

src/components/constants/display-sections.js
```diff
@@ -1 +1,3 @@
 export const CHALLENGES_SCREEN = 'Challenges';
+export const DUMMY_PROFILE = 'dummyProfile.png';
+export const IMAGE_URL = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL;
```
swarajpure, line 3: This seems like a long way to do a simple thing 😅 
Import something from env variables in one file, export it, re-import again in the same file without doing anything with it.

We're sort of re-declaring it, which is not really necessary here. What do you think?

I was saying that in the other file itself you can define it as `const IMAGE_URL = process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL` and go ahead

src/components/idleMembers/card/index.tsx
swarajpure: Great. Thanks for the clarification! Looks good then! Also could you add screenshot of properly working images?

src/components/constants/display-sections.js
abhinav1590: You are right😅, I will make the changes right away

src/components/idleMembers/card/index.tsx
abhinav1590: ![image](https://user-images.githubusercontent.com/86315672/132954190-89761c7a-a96e-40f3-b589-17db80e40e1f.

swarajpure: Cool!

swarajpure (approved):
Looks good to me!

Great job!

sumitd94 (approved):

PR #199, opened 2021-09-10
Added the nav bar similar to home page

### LINKED ISSUE #145: Create navbar
[Requirements](https://github.com/Real-Dev-Squad/website-todo-items/issues/29)

rohan09-raj:
Closes #145 
Before :
Screenshot (4)
After :
(Mobile View)
Screenshot (19)
(Tablet View)
Screenshot (18)
(Desktop/Laptop View)
Screenshot (17)

### DIFF
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -55,8 +55,8 @@ const Layout: FC<Props> = ({ children }) => {
         }
         </div>
         {children}
-        <Footer />
       </div>
+      <Footer />
     </div>
   );
 };
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -3,7 +3,7 @@ import Link from 'next/link';
 import { useRouter } from 'next/router';
 import Footer from '@/components/footer';
 import styles from '@/components/Layout/Layout.module.scss';
-import NavBar from '../navBar';
+import NavBar from '@/components/navBar';
 
 interface Props {
   children?: ReactNode
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -31,19 +31,20 @@ const Layout: FC<Props> = ({ children }) => {
   return (
     <div className={styles.layout}>
       <NavBar />
-      <div className={styles.header}>
-        {navBarContent('Tasks', '/')}
-        |
-        {navBarContent('Mine', '/mine')}
-        |
-        {navBarContent('DS', '/challenges')}
-        |
-        {navBarContent('Open PRs', '/openPRs')}
-        |
-        {navBarContent('Stale PRs', '/stale-pr')}
-        |
-        {navBarContent('Idle Members', '/idle-members')}
-        {
+      <div className={styles.wrapper}>
+        <div className={styles.header}>
+          {navBarContent('Tasks', '/')}
+          |
+          {navBarContent('Mine', '/mine')}
+          |
+          {navBarContent('DS', '/challenges')}
+          |
+          {navBarContent('Open PRs', '/openPRs')}
+          |
+          {navBarContent('Stale PRs', '/stale-pr')}
+          |
+          {navBarContent('Idle Members', '/idle-members')}
+          {
           (dev)
             && (
               <>
@@ -52,9 +53,10 @@ const Layout: FC<Props> = ({ children }) => {
               </>
             )
         }
+        </div>
+        {children}
+        <Footer />
       </div>
-      {children}
-      <Footer />
     </div>
   );
 };
--- a/src/components/Layout/index.tsx
+++ b/src/components/Layout/index.tsx
@@ -1,9 +1,9 @@
 import { FC, ReactNode } from 'react';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import Image from 'next/image';
 import Footer from '@/components/footer';
 import styles from '@/components/Layout/Layout.module.scss';
+import NavBar from '../navBar';
 
 interface Props {
   children?: ReactNode
@@ -30,14 +30,7 @@ const Layout: FC<Props> = ({ children }) => {
 
   return (
     <div className={styles.layout}>
-      <div className={styles.rdsHeader}>
-        <a href="https://realdevsquad.com" target="_blank" rel="noreferrer">
-          <Image width="45px" height="45px" className={styles.logo} src="/RDSLogo.png" alt="real-dev squad" />
-        </a>
-        <Link href="/">
-          <div className={styles.heading}>Status</div>
-        </Link>
-      </div>
+      <NavBar />
       <div className={styles.header}>
         {navBarContent('Tasks', '/')}
         |
--- a/src/components/navBar/index.tsx
+++ b/src/components/navBar/index.tsx
@@ -9,7 +9,6 @@ const NavBar = () => (
       <li>
         <a className={styles.logo} href="https://realdevsquad.com"><Image width="45px" height="45px" src={RDSLogo} alt="real-dev squad" /></a>
       </li>
-      <li><a href="https://realdevsquad.com">Home</a></li>
       <li><a href="https://welcome.realdevsquad.com/">Welcome</a></li>
       <li><a href="https://www.realdevsquad.com/events.html">Events</a></li>
       <li><a href="https://members.realdevsquad.com/">Members</a></li>
--- a/src/components/navBar/index.tsx
+++ b/src/components/navBar/index.tsx
@@ -1,0 +1,24 @@
+import styles from '@/components/navBar/navBar.module.scss';
+import Image from 'next/image';
+
+const RDSLogo = '/RDSLogo.png';
+
+const NavBar = () => (
+  <div>
+    <nav className={styles.navBar}>
+      <ul>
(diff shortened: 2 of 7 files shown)


Pavangbhat (changes requested):

src/components/navBar/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import Image from 'next/image';
+
+const RDSLogo = '/RDSLogo.png';
+
+const NavBar = () => (
+  <div>
+    <nav className={styles.navBar}>
+      <ul>
+        <li style={{ padding: '7px' }}>
+          <Image width="45px" height="45px" src={RDSLogo} alt="real-dev squad" />
```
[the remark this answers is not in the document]
rohan09-raj: Yes, I have made the requested changes

src/components/navBar/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <ul>
+        <li>
+          <a style={{ padding: '7px' }} href="https://realdevsquad.com"><Image width="45px" height="45px" src={RDSLogo} alt="real-dev squad" /></a>
+        </li>
+        <li><a href="https://realdevsquad.com">Home</a></li>
+        <li><a href="https://welcome.realdevsquad.com/">Welcome</a></li>
+        <li><a href="https://www.realdevsquad.com/events.html">Events</a></li>
+        <li><a href="https://members.realdevsquad.com/">Members</a></li>
+        <li><a href="https://crypto.realdevsquad.com/">Crypto</a></li>
+        <li><a style={{ color: '#87d870' }} href="https://status.realdevsquad.com/">Status</a></li>
```
Pavangbhat, line 18: **Suggestion:** Better to avoid inline styles to maintain consistency. Apart from that everything looks good to me :+1: 
Nice work :tada:

Pavangbhat (approved):

src/components/navBar/index.tsx
Pavangbhat, line 18: Suggestion: Better to avoid inline styles to maintain consistency. Apart from that everything looks good to me :+1:
Nice work :tada:

rohan09-raj: Thanks😀
Yeah, I tried to avoid inline styles but due to low specificity, the styles can't be applied.

swarajpure: In that case, we could have an id for this element and style it in css file?

rohan09-raj: I tried with Id too but the css was not applying.

src/components/Layout/index.tsx
```diff
@@ -1,9 +1,9 @@
 import { FC, ReactNode } from 'react';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import Image from 'next/image';
 import Footer from '@/components/footer';
 import styles from '@/components/Layout/Layout.module.scss';
+import NavBar from '../navBar';
```
whyDontI, line 6: Please use the absolute imports here, refer above imports. 
https://nextjs.org/docs/advanced-features/module-path-aliases

src/components/navBar/index.tsx
whyDontI: We can always use CSS classes, any particular reason for not using them?
Give it a class as `active` and move this style to the SCSS file.

src/components/navBar/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import Image from 'next/image';
+
+const RDSLogo = '/RDSLogo.png';
+
+const NavBar = () => (
+  <div>
+    <nav className={styles.navBar}>
+      <ul>
+        <li>
+          <a style={{ padding: '7px' }} href="https://realdevsquad.com"><Image width="45px" height="45px" src={RDSLogo} alt="real-dev squad" /></a>
```
whyDontI, line 11: Make a separate class for this too, and move the styles in SCSS file

src/components/navBar/navBar.module.scss
```diff
@@ -0,0 +1,31 @@
...
+    color: #ffffff;
+    font-weight: bolder;
+    text-align: center;
+    padding: 19px 16px;
+    text-decoration: none;
+}
+
+.navBar li a:hover {
+    background-color: rgba(255, 255, 255, 0.2);
+}
```
whyDontI, line 31: Use 2 space indentation please

whyDontI (changes requested):
Requested some changes 😄

src/components/navBar/index.tsx
rohan09-raj: Yeah I tried but I don't know why the styles were not applying using classes or ids.

rohan09-raj: The issue for the css not being applied was that it is being striken out and was not overriding the original white colour of the text which was applied to all list elements.

Solution : Using this -> 

.navBar li .active {}

Instead of this ->

.active {}
Screenshot (4)

src/components/navBar/index.tsx
```diff
@@ -0,0 +1,24 @@
+import styles from '@/components/navBar/navBar.module.scss';
+import Image from 'next/image';
+
+const RDSLogo = '/RDSLogo.png';
+
+const NavBar = () => (
+  <div>
```
whyDontI, line 7: We don't need this `div` can we drop it?

src/components/navBar/navBar.module.scss
```diff
@@ -0,0 +1,39 @@
+.navBar {
+  width: 100vw;
+  font-size: 1rem;
+  margin: 0 -2em;
```
whyDontI, line 4: I don't think using negative margins is a good idea.
Let's try to solve the problem from its root.

In this case, we can fix this problem by removing margins from `html` styles, and we won't need the margin in navbar.

https://github.com/Real-Dev-Squad/website-status/blob/6a25cf74187cf413b93418f5a0fab62c1545383d/src/styles/index.scss#L7

src/components/navBar/navBar.module.scss
```diff
@@ -0,0 +1,39 @@
+.navBar {
+  width: 100vw;
+  font-size: 1rem;
```
whyDontI, line 3: The `font-size` is by default `1rem` no need to mention it specifically.
https://github.com/Real-Dev-Squad/website-status/blob/6a25cf74187cf413b93418f5a0fab62c1545383d/src/styles/index.scss#L17

whyDontI (changes requested):

src/components/navBar/index.tsx
rohan09-raj: Yes we can drop it.

src/components/navBar/navBar.module.scss
rohan09-raj: Yes you are right, but in that case we have to add margins for rest of the UI except navBars (for the mobile view). So I tried it wrapping the rest content in a div and add margins but it was creating some problems. I will try to find a way to do this and as of now I have to redesign (some changes are required only) navBar as it has to look same all across the websites.

rahilkapoor:
Looks perfect to me!
But there is a slight change in the navbar...We have removed the home tab thus we will be navigating to the home page via the logo itself

rohan09-raj:
> Looks perfect to me!
> But there is a slight change in the navbar...We have removed the home tab thus we will be navigating to the home page via the logo itself
> navbar
> 
> Ignore the right corner sign-in thing for now!

Ohh okay, I forgot to remove it, will do it ASAP. And Is the padding fine or should I increase it to match the design.

rahilkapoor:
> > Looks perfect to me!
> > But there is a slight change in the navbar...We have removed the home tab thus we will be navigating to the home page via the logo itself
> > navbar
> > Ignore the right corner sign-in thing for now!
> 
> Ohh okay, I forgot to remove it, will do it ASAP. And Is the padding fine or should I increase it to match the design.

You might have to increase the padding for it to look something similar...(make changes which you feel are necessary to make it look alike)✔

whyDontI (changes requested):
Navbar is causing horizontal scroll, can we fix that, please?

rohan09-raj:
@whyDontI I have tested on multiple screen sizes and none of them is causing is issue. Can you please share the dimensions in which the horizontal scroll is caused or some reference please.

whyDontI:
Not a problem with resolution, the problem is with the `tasks` page I guess, it's not replicating on any other pages.
Can you check?

rohan09-raj:
> Not a problem with resolution, the problem is with the `tasks` page I guess, it's not replicating on any other pages.
> Can you check?



> Not a problem with resolution, the problem is with the `tasks` page I guess, it's not replicating on any other pages.
> Can you check?

Unable to get you. 😅😅, `tasks` page is working fine with other pages too. (`mine`, `DS`, `Open PRs` etc.)

rohan09-raj:
> Not a problem with resolution, the problem is with the `tasks` page I guess, it's not replicating on any other pages.
> Can you check?

Made the suggested change. 
- Navbar `width: 100vw` to `width: 100%`
This was causing the navBar to scroll when data is loaded.

whyDontI (approved):
Good work 🥳

rohan09-raj:
> Good work 🥳

Thank you 😃

PR #200, opened 2021-09-14
Edit Tasks from the site

rohandhar1234:
I have made changes in two files. In the tasks\card index.tsx:

- I have initialized the value of task contents to a new variable cardDetails
- function handleTitleChange(event, changedDetails) will trigger the event i.e whenever enter key is pressed it will store the cardContent in a variable toChange.  toChange.changedDetails will be the field where we make changes in the front-end. Now, function stripHtml will run on the edits that we do on the site and will remove all the html tags such as <br>
- After this onContentChange(toChange) will be called. This is a props that I have used and will be later used to fetch api
- Finally, I have passed contentEditable and event listener to the <span> fields of all the editable contents that we do on site( title, assignee etc)

In the pages section, I have made changes in the index.tsx file

- When I call <Card>, I pass the properties and set the values of shouldEdit to edit (which can be true or false) and then I set the value of onContentChange to updateCardContent(newDetails) which will change values of desired field and fetch the api in the updateCardContent(cardDetails) function.
- I have also used router here to pass the query parameter of ?edit=true and declared a value of this parameter to a variable with name edit and then later passed this value in the renderCarList function which is used for functioning only when shouldEdit= True.

### DIFF
--- a/src/components/tasks/card/index.tsx
+++ b/src/components/tasks/card/index.tsx
@@ -1,62 +1,92 @@
-import { FC, useState } from 'react';
-import classNames from '@/components/tasks/card/card.module.scss';
-import task from '@/interfaces/task.type';
-import Image from 'next/image';
+import { FC, useState, useEffect } from "react";
+import classNames from "@/components/tasks/card/card.module.scss";
+import task from "@/interfaces/task.type";
+import Image from "next/image";
 import {
   ACTIVE,
   ASSIGNED,
   COMPLETED,
-  UNASSIGNED,
-} from '@/components/constants/task-status';
+} from "@/components/constants/task-status";
 
-const moment = require('moment');
+const moment = require("moment");
 
 type Props = {
   content: task;
 };
 
-const Card: FC<Props> = ({ content }) => {
-  const {
-    title, endsOn, startedOn, status, assignee,
-  } = content;
+const Card: FC<Props> = ({ content, shouldEdit, onContentChange }) => {
+  // const {
+  //   id, title, endsOn, startedOn, status, assignee,
+  // } = content;
+  const cardDetails = (content);
 
-  const [assigneeProfilePic, setAssigneeProfilePic] = useState(`${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${assignee}/img.png`);
-  const contributorImageOnError = () => setAssigneeProfilePic('dummyProfile.png');
+  const [assigneeProfilePic, setAssigneeProfilePic] = useState(
+    `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${cardDetails.assignee}/img.png`
+  );
+  const contributorImageOnError = () =>
+    setAssigneeProfilePic("dummyProfile.png");
 
-  const localStartedOn = new Date(parseInt(startedOn, 10) * 1000);
+  const localStartedOn = new Date(parseInt(cardDetails.startedOn, 10) * 1000);
   const fromNowStartedOn = moment(localStartedOn).fromNow();
 
-  const localEndsOn = new Date(parseInt(endsOn, 10) * 1000);
+  const localEndsOn = new Date(parseInt(cardDetails.endsOn, 10) * 1000);
   const fromNowEndsOn = moment(localEndsOn).fromNow();
 
-  const statusFontColor = status === ACTIVE || ASSIGNED || COMPLETED ? '#00a337' : '#f83535';
+  const statusFontColor =
+    status === ACTIVE || ASSIGNED || COMPLETED ? "#00a337" : "#f83535";
 
-  const iconHeight = '25px';
-  const iconWidth = '25px';
+  const iconHeight = "25px";
+  const iconWidth = "25px";
 
   const cardClassNames = [classNames.card];
 
   function isTaskOverdue() {
     const currentDate = new Date();
     const timeLeft = localEndsOn.valueOf() - currentDate.valueOf();
-    return status !== COMPLETED && status !== UNASSIGNED && timeLeft <= 0;
+    return status !== COMPLETED && timeLeft <= 0;
+  }
+
+  function stripHtml(html) {
+    let tmp = document.createElement("DIV");
+    tmp.innerHTML = html;
+    return tmp.textContent || tmp.innerText || "";
+  }
+
+  function handleTitleChange(event, changedDetails) {
+    if (event.key === "Enter") {
+      const toChange = cardDetails;
+      toChange.[changedDetails]= stripHtml(event.target.innerHTML);
+      onContentChange(toChange);
+    }
   }
 
   if (isTaskOverdue()) {
     cardClassNames.push(classNames.overdueTask);
   }
   return (
-    <div className={`
+    <div
+      className={`
     ${classNames.card}
     ${isTaskOverdue() && classNames.overdueTask}
 `}
     >
       <div className={classNames.cardItems}>
-        <span className={classNames.cardTitle}>{title}</span>
+        <span
+          className={classNames.cardTitle}
+          contentEditable={shouldEdit}
+          onKeyPress={(e) => handleTitleChange(e, 'title')}
+        >
+          {cardDetails.title}
+        </span>
         <span>
           <span className={classNames.cardSpecialFont}>Status:</span>
-          <span className={classNames.cardStatusFont} style={{ color: statusFontColor }}>
-            {status}
+          <span
+            className={classNames.cardStatusFont}
+            contentEditable={shouldEdit}
+            onKeyPress={(e) => handleTitleChange(e, 'status')}
+            style={{ color: statusFontColor }}
+          >
+            {cardDetails.status}
           </span>
         </span>
       </div>
@@ -68,21 +98,25 @@ const Card: FC<Props> = ({ content }) => {
             width={iconWidth}
             height={iconHeight}
           />
-          <span className={classNames.cardSpecialFont}>
-            Due Date
-          </span>
-          <span className={classNames.cardStrongFont}>{fromNowEndsOn}</span>
+          <span className={classNames.cardSpecialFont}>Due Date</span>
+          <span className={classNames.cardStrongFont}
+          contentEditable={shouldEdit}
+          onKeyPress={(e) => handleTitleChange(e, 'fromNowEndsOn')}>{fromNowEndsOn}</span>
         </span>
       </div>
       <div className={classNames.cardItems}>
-        <span className={classNames.cardSpecialFont}>
-          Started
-          {' '}
-          {fromNowStartedOn}
+        <span className={classNames.cardSpecialFont}
+        contentEditable={shouldEdit}
+        onKeyPress={(e) => handleTitleChange(e, 'fromNowStartedOn')}>
+          Started {fromNowStartedOn}
         </span>
         <span>
           <span className={classNames.cardSpecialFont}>Assignee:</span>
-          <span className={classNames.cardStrongFont}>{assignee}</span>
+          <span className={classNames.cardStrongFont}
+          contentEditable={shouldEdit}
+          onKeyPress={(e) => handleTitleChange(e, 'assignee')}>
+            {cardDetails.assignee}
+          </span>
           <span>
             <img
               className={classNames.contributorImage}
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -6,14 +6,49 @@ import useFetch from '@/hooks/useFetch';
 import classNames from '@/styles/tasks.module.scss';
 import task from '@/interfaces/task.type';
 import Accordion from '@/components/Accordion';
+import axios from 'axios';
+import { useRouter } from 'next/router';
 
 const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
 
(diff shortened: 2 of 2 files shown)


rohandhar1234:
In the newer commits, I have passed the default props to <Card> in the files it was called. These props are responsible for task editing.
I have also properly added the query parameter "?edit=true" and now only those users will be able to edit tasks which have both super_user and admin rights. As Ankush has both rights and to what I believe only he has both of these rights so only Ankush will be able to edit tasks when the query parameter is entered.

src/components/tasks/card/index.tsx
```diff
@@ -39,24 +46,81 @@ const Card: FC<Props> = ({ content }) => {
...
+      toChange.startedOn = stripHtml(event.target.innerHTML);
+      onContentChange(toChange);
+    }
+  }
+  function handleAssigneeChange(event: any) {
+    if (event.key === 'Enter') {
+      const toChange = cardDetails;
+      toChange.assignee = stripHtml(event.target.innerHTML);
+      onContentChange(toChange);
+    }
```
ankushdharkar, line 91: Lot of repeated code here. Can we reuse a single function somehow?

ankushdharkar (changes requested):

src/pages/index.tsx
```diff
@@ -8,21 +10,54 @@ import task from '@/interfaces/task.type';
 import Accordion from '@/components/Accordion';
 
 const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
+const SELF_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/users/self`;
 
-function renderCardList(tasks: task[]) {
-  return tasks.map((item: task) => <Card content={item} key={item.id} />);
+async function updateCardContent(cardDetails: any) {
```
whyDontI, line 15: You can import the type/interface for tasks from [here](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/interfaces/task.type.ts)

src/pages/index.tsx
```diff
@@ -8,21 +10,54 @@ import task from '@/interfaces/task.type';
...
+  let response = {};
+  let error = {};
+  try {
+    response = await axios({
+      method: 'patch',
+      url: `https://api.realdevsquad.com/tasks/${cardDetails.id}`,
```
whyDontI, line 21: Use the URL variables defined on line no. 12 & 13

src/pages/index.tsx
```diff
@@ -8,21 +10,54 @@ import task from '@/interfaces/task.type';
...
+      data: cardDetails,
+      headers: {
+        'Content-type': 'application/json',
+      },
+      withCredentials: true,
+    });
+  } catch (err:any) {
+    error = err;
+    // eslint-disable-next-line no-console
+    console.log('Error', error);
```
whyDontI, line 32: Show toast on errors
https://github.com/Real-Dev-Squad/website-status/blob/develop/src/helperFunctions/toast.ts

src/pages/index.tsx
```diff
@@ -8,21 +10,54 @@ import task from '@/interfaces/task.type';
 import Accordion from '@/components/Accordion';
 
 const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
+const SELF_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/users/self`;
 
-function renderCardList(tasks: task[]) {
-  return tasks.map((item: task) => <Card content={item} key={item.id} />);
+async function updateCardContent(cardDetails: any) {
+  let response = {};
+  let error = {};
```
whyDontI, line 17: Make it `const` if possible

src/pages/index.tsx
```diff
@@ -8,21 +10,54 @@ import task from '@/interfaces/task.type';
...
+  try {
+    response = await axios({
```
whyDontI, line 19: Use our own [fetch helper function](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/helperFunctions/fetch.ts)

src/pages/index.tsx
```diff
@@ -8,21 +10,54 @@ import task from '@/interfaces/task.type';
...
 const Index: FC = () => {
+  const router = useRouter();
+  const { query } = router;
   let tasks: task[] = [];
   const [filteredTask, setFilteredTask] = useState<any>([]);
-
   const { response, error, isLoading } = useFetch(TASKS_URL);
-
+  const [roles, setRoles] = useState(false);
+  const edit = !!query.edit && roles;
```
whyDontI, line 56: Can you explain this?

src/pages/index.tsx
```diff
@@ -35,6 +70,23 @@ const Index: FC = () => {
     }
   }, [isLoading, response]);
 
+  useEffect(() => {
+    const url = SELF_URL;
+    const fetchData = async () => {
+      try {
+        const r_esponse = await fetch(url);
+        const json = await r_esponse.json();
+        if (json.user.roles.admin && json.user.roles.super_user === true) {
```
whyDontI, line 79: I believe you need OR here.
Also, we don't need `===` for boolean values

src/pages/index.tsx
```diff
@@ -35,6 +70,23 @@ const Index: FC = () => {
     }
   }, [isLoading, response]);
 
+  useEffect(() => {
+    const url = SELF_URL;
+    const fetchData = async () => {
+      try {
+        const r_esponse = await fetch(url);
+        const json = await r_esponse.json();
```
whyDontI, line 78: Use our own [fetch helper function](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/helperFunctions/fetch.ts)

whyDontI (changes requested):

src/pages/index.tsx
rohandhar1234: We can use OR here if we want user to be either admin or super_user but I used AND to include only those users which have both. Thinking only Ankush has both rights, I used AND. Should I change it to OR

src/pages/index.tsx
rohandhar1234: I used the state variable roles with initial value FALSE and then used it to initialize the edit(variable). That is, if query parameter is true and roles is also true, only then the user will be able to fill in and do edits

src/pages/index.tsx
rohandhar1234: Do you mean like this?
```async function updateCardContent(cardDetails: task) {
    const responsePromise = await axios({
      method: 'patch',
      url: `https://api.realdevsquad.com/tasks/${cardDetails.id}`,
      params: null,
      data: cardDetails,
      headers: {
        'Content-type': 'application/json',
      },
      withCredentials: true,
    });
    return responsePromise;

Comment (rohandhar1234) [reply src/pages/index.tsx:19]: Do you mean like this?
```async function updateCardContent(cardDetails: task) {
  try {
    const response = await fetch({
      url: `${TASKS_URL}/${cardDetails.id}`,
      method: 'patch',
      params: null,
      data: cardDetails,
      headers: {
        'Content-type': 'application/json',
      },
    });
    // eslint-disable-next-line no-console
    console.log('Response', response);
  } catch (err:any) {
    // eslint-disable-next-line no-console
    console.log(toast);
  }
}

File: src/pages/index.tsx (hunk 8 above)
Comment (rohandhar1234) [reply src/pages/index.tsx:79]: About === the code shows eslint issue whenever I try to use ==.
I know that === is used for type checking and value checking and == is just to check whether values are equal but it shows 

Expected '===' and instead saw '=='.eslinteqeqeq

File: src/pages/index.tsx (hunk 9 above)
Comment (rohandhar1234) [reply src/pages/index.tsx:78]: Is this the right way?
```useEffect(() => {
    const url = SELF_URL;
    const fetchData = async () => {
      try {
        const r_esponse = await fetch({ url });
        const adminUser :boolean = r_esponse.data.user.roles.admin;
        const superUser :boolean = r_esponse.data.user.roles.super_user;
        if (adminUser || superUser === true) {
          setRoles(true);
        }
      } catch (error_) {
        // eslint-disable-next-line no-console
        console.log('error', error_);
      }
    };

    fetchData();
  }, []);

File: src/pages/index.tsx (hunk 7 above)
Comment (whyDontI) [reply src/pages/index.tsx:61]: Got it 👍🏽 
You need to name the variables with suitable names.
Think of a better name like `isUserAuthorized` or `isAdmin` or `isEditable`.

File: src/pages/index.tsx (hunk 8 above)
Comment (whyDontI) [reply src/pages/index.tsx]: Loose the `if` and make it `setRoles(adminUser || superUser)`

Comment (whyDontI) [reply src/pages/index.tsx]: Also, I don't see any `roles` field [here](https://github.com/Real-Dev-Squad/website-api-contracts/tree/main/users#user-object) and [here](https://github.com/Real-Dev-Squad/website-backend/blob/605fcec70157a760a165fb212f22758c9e0fd2b5/docs/swaggerDefinition.js#L412) where did you find the `roles` field.

Comment (rohandhar1234) [reply src/pages/index.tsx]: I used the SELF_URL and fetched it. This gives the user details (username, GitHu_id etc and roles also)

File: src/pages/index.tsx (hunk 9 above)
Comment (whyDontI) [reply src/pages/index.tsx]: Use [object destructuring](https://javascript.info/destructuring-assignment#object-destructuring)

Comment (whyDontI) [reply src/pages/index.tsx]: Any reason for `const url = SELF_URL`?

File: src/pages/index.tsx (hunk 8 above)
Comment (whyDontI) [reply src/pages/index.tsx]: We have API-Contracts and that should be up to date; in case it's not, you should make the concerned person / AO aware of it.
I'll make an issue to update the API Contracts 👍🏽

File: src/pages/index.tsx (hunk 9 above)
Comment (rohandhar1234) [reply src/pages/index.tsx]: I initialized this url globally and thought it would be better to use it this way rather than writing the whole url again.
I could just use it without creating another variable also. I will do that

Comment (whyDontI) [reply src/pages/index.tsx]: Also what is with this `snake_case` variable `r_esponse`?
We strictly use `camelCase` for our variables.

Comment (rohandhar1234) [reply src/pages/index.tsx]: Alright I will sort it out

File: src/pages/index.tsx (hunk 8 above)
Comment (rohandhar1234) [reply src/pages/index.tsx]: Okay

File: src/pages/index.tsx [hunk 10]
```diff
@@ -15,21 +18,51 @@ import {
 } from '@/components/constants/task-status';
 
 const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
+const SELF_URL = 'https://api.realdevsquad.com/users/self';
```
swarajpure, line 21: Why not use `NEXT_PUBLIC_BASE_URL` here?

src/pages/index.tsx
whyDontI: Check how we are using this helper function in other places in the code base

src/pages/index.tsx
```diff
@@ -44,6 +77,25 @@ const Index: FC = () => {
...
   }, [isLoading, response]);
 
+  useEffect(() => {
+    const url = SELF_URL;
+    const fetchData = async () => {
+      try {
+        const r_esponse = fetch({ url });
+        const fetchPromise = await r_esponse.requestPromise;
+        const adminUser :boolean = fetchPromise.data.user.roles.admin;
+        const superUser :boolean = fetchPromise.data.user.roles.super_user;
```
swarajpure, line 87: Better to use optional chaining here?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/pages/index.tsx
whyDontI: Check how we are using this helper function in other places in the code base

src/pages/index.tsx
rohandhar1234: The NEXT_PUBLIC_BASE_URL is the staging url and has no roles with admin or super_user. The staging url only has member roles defined there

src/pages/index.tsx
rohandhar1234: Alright I will go through this and see if it fits the code.

rohandhar1234: Yes this is useful. Thank you

src/components/tasks/card/index.tsx
```diff
@@ -13,24 +13,31 @@ const moment = require('moment');
 
 type Props = {
   content: task;
+  shouldEdit: boolean;
```
swarajpure, line 16: how about `isEditable`?

src/pages/index.tsx
swarajpure: Not using it?

rohandhar1234: I am going to use it and update the pr

src/components/tasks/card/index.tsx
rohandhar1234: Do you want me to change it's name to something else?
I have already used isEditable for the 'edit' variable and isUserAuthorised for the state variable 'role'
It will look like:
shouldEdit= isEditable;
where,
isEditable = !!query.edit && isUserAuthorised;

If you want me to change it's name then I will look for something better

rohandhar1234: isUserAuthorised will tell us if user has admin rights or not and isEditable will only be 'true' if the query parameter will be  'edit=true' and the user who's logged in has admin roles.

PR #217, opened 2021-11-02
Fix unwanted edit task request for unauthorized user

### LINKED ISSUE #212: Patch request should not happen if the user is not Authorized to edit tasks
# Problem
On the https://status.realdevsquad.com/?edit=true page, the user without `super_user` privileges cannot edit the content, but upon hitting enter the browser makes a PATCH request.

Ideally, this request should not be made for any user other than `super_user`

shubham-y:
Fixes #212

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -22,21 +22,23 @@ const TASKS_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/tasks`;
 const SELF_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/users/self`;
 const STATUS_ORDER = [ACTIVE, ASSIGNED, COMPLETED, PENDING, UNASSIGNED];
 
-async function updateCardContent(id: string, cardDetails: task) {
-  try {
-    const { requestPromise } = fetch({
-      url: `${TASKS_URL}/${id}`,
-      method: 'patch',
-      data: cardDetails,
-    });
-    await requestPromise;
-    toast(SUCCESS, 'Changes have been saved !');
-  } catch (err: any) {
-    if ('response' in err) {
-      toast(ERROR, err.response.data.message);
-      return;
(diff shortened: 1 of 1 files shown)


src/pages/index.tsx
```diff
@@ -87,7 +91,7 @@ const Index: FC = () => {
           superUser: data.roles?.super_user,
         };
         const { adminUser, superUser } = userRoles;
-        setIsUserAuthorized(adminUser || superUser);
+        setIsUserAuthorized(!!adminUser || !!superUser);
```
harshith-venkatesh, line 94: Good usage of `!!` as we might expect `undefined` for the roles here

src/pages/index.tsx
```diff
@@ -46,7 +48,9 @@ function renderCardList(tasks: task[], isEditable: boolean) {
       content={item}
       key={item.id}
       shouldEdit={isEditable}
-      onContentChange={async (id: string, newDetails: any) => updateCardContent(id, newDetails)}
+      onContentChange={async (id: string, newDetails: any) => {
+        updateCardContent(id, newDetails, isEditable);
```
harshith-venkatesh, line 52: What is the significance of `{}` here?

harshith-venkatesh: NIT: Could we think of conditionally calling the function instead of passing it as argument and then checking the condition?

harshith-venkatesh: What is your opinion on `isEditable && updateCardContent(params)`, would this be better than passing it as params and checking the condition?

harshith-venkatesh, line 52: What is your opinion on `isEditable && updateCardContent(params)` , would this be better than passing it as params and checking the condition?

shubham-y: Yes, this way will be better than the current one.
I will do the changes

shubham-y: @harshith-venkatesh Sorry the branch got deleted. I have raised a new PR #221

PR #221, opened 2021-11-06
Fix unwanted edit task request for unauthorized user

### LINKED ISSUE #212: Patch request should not happen if the user is not Authorized to edit tasks

shubham-y:
Fixes #212


harshith-venkatesh (approved):
Great work👍

PR #228, opened 2021-11-20
Fix/navbar - Replaced media queries

jjagannn:
Usage of media queries have been replaced with direct fixes and variables


jjagannn:
Need to remove this PR

Issue #223, opened 2021-11-12
Maintain the order of tasks/idle-members after a card is dropped on other


Pavangbhat:
Availability panel allows reordering tasks/idle-member card as per priority and once a card is dropped on other a PATCH request is made and re-rendering of the component takes place with new data and the order is no longer maintained 
reference :point_down: 

https://user-images.githubusercontent.com/45538392/141510864-2034e5e1-c751-43d5-a5a1-b56a76ac6824.mp4

PR #236, opened 2021-12-30
Stop reordering

akshay1502:
closes :- #223


Pavangbhat (changes requested):
The order is maintained but on drop of a task if its successfully assigned to a member it should pop out of the list since its no more unassigned task
And should fetch the new data keep our task/idle members list updated

akshay1502:
@Pavangbhat after dropping task on successful assignation task status becomes assigned, as we are only showing unassigned task on page so any how that task gets out of list.

Pavangbhat:
> 

It gets out of the list once we refresh the page till then it is going to be present on the list, that is what I found when I ran the changes that you have made 🤔

akshay1502:
How?? after each drop action we are fetching new data, so the current task being dropped will be marked as assigned, and will not be displayed in the new list. No need for page refresh.

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -34,8 +34,13 @@ const DragDropcontext: FC<dragDropProps> = ({
   const [isTaskOnDrag, setIsTaskOnDrag] = useState<boolean>(false);
 
   useEffect(() => {
-    setTaskList(unAssignedTasks);
-    setMemberList(idleMembers);
+    const oldTasksIds = taskList.map((Task: task) => Task.id);
+    const newTasks = unAssignedTasks.filter((Task: task) => !oldTasksIds.includes(Task.id));
+    const newTaskList = [...taskList, ...newTasks];
```
whyDontI, line 39: Let's move this logic to a function

whyDontI (changes requested):

Pavangbhat:
> How?? after each drop action we are fetching new data, so the current task being dropped will be marked as assigned, and will not be displayed in the new list. No need for page refresh. @Pavangbhat

Sorry, @akshay1502 I think I have tested it new task workflow so it behaved like that, let me test it with the old task status.

Pavangbhat (commented):
Everything looks good to me 😃 
Please make changes suggested by Nikhil 
Great Work 🎉

src/components/availability-panel/drag-drop-context/index.tsx
Pavangbhat: @akshay1502 Let's move this logic into a separate function

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -34,9 +34,11 @@ const DragDropcontext: FC<dragDropProps> = ({
   const [isTaskOnDrag, setIsTaskOnDrag] = useState<boolean>(false);
 
   useEffect(() => {
-    const oldTasksIds = taskList.map((Task: task) => Task.id);
-    const newTasks = unAssignedTasks.filter((Task: task) => !oldTasksIds.includes(Task.id));
-    const newTaskList = [...taskList, ...newTasks];
+    const newTaskList = ():task[] => {
```
Pavangbhat, line 37: We should place it in the **helperFunctions** folder  
If possible lets have better name such that by looking at it we should know where we will be using it

Pavangbhat (changes requested):
Other than the changes requested everything looks good to me

src/components/availability-panel/drag-drop-context/index.tsx
akshay1502: Any suggestions for the function name @Pavangbhat.

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -41,8 +42,10 @@ const DragDropContextWrapper: FC<dragDropProps> = ({
   const [isTaskOnDrag, setIsTaskOnDrag] = useState<boolean>(false);
 
   useEffect(() => {
-    setTaskList(unAssignedTasks);
-    setMemberList(idleMembers);
+    setTaskList(appendNewTasks(taskList, unAssignedTasks));
+    const newMembers = idleMembers.filter((member) => !memberList.includes(member));
+    const newMemberList = [...memberList, ...newMembers];
```
whyDontI, line 47: Can we put this logic in a function too, just like we did tasks in `appendNewTasks`

src/helperFunctions/getTasks.ts
```diff
@@ -0,0 +1,10 @@
+import task from '@/interfaces/task.type';
+
+export const appendNewTasks = (
+  taskList: task[],
+  unAssignedTasks: task[]
+  ) :task[] => {
+    const oldTasksIds = taskList.map((Task: task) => Task.id);
+    const newTasks = unAssignedTasks.filter((Task: task) => !oldTasksIds.includes(Task.id));
+    return [...taskList, ...newTasks];
+};
```
whyDontI, line 10: Let's move this function in the component itself.
Helper functions are the functions that we'll be using more often in multiple places.

src/mocks/idlemembers.json
```diff
@@ -0,0 +1,5 @@
+{
+  "idleMembers": {
+    "idleMemberUserNames": ["rohan-rajgupta", "sumit", "rohit", "tanya", "akshay", "shubham", "devashish", "lakshay", "rucha", "swebert", "pavan", "nikhil", "ankur", "swaraj"]
+  }
+}
```
whyDontI, line 5: We have added API response mocking in the codebase, please refer to [this](https://mswjs.io/docs/api/rest) doc to learn to write the handler for the same.

Check [this message on discord](https://discord.com/channels/673083527624916993/741616265050849282/937589378593726495).
Please go through [CONTRIBUTING.md](https://github.com/Real-Dev-Squad/website-status/blob/develop/CONTRIBUTING.md#to-run-the-project-with-mocked-apis)

whyDontI (changes requested):
Good work 🥳 
Requesting little changes

src/components/availability-panel/drag-drop-context/index.tsx
akshay1502: done.

yarn.lock
```diff
@@ -23,9 +23,9 @@
...
 
 "@babel/compat-data@^7.17.7", "@babel/compat-data@^7.18.8":
-  version "7.18.8"
-  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.18.8.tgz#2483f565faca607b8535590e84e7de323f27764d"
-  integrity sha512-HSmX4WZPPK3FUxYp7g2T6EyO8j96HlZJlxmKPSh6KAcqwyDrfx7hKjXpAW/0FhFfTJsR0Yt4lAjLI2coMptIHQ==
+  version "7.18.13"
+  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.18.13.tgz#6aff7b350a1e8c3e40b029e46cbe78e24a913483"
+  integrity sha512-5yUzC5LqyTFp2HLmDoxGQelcdYgSpP9xsnMWBphAscOdFrHSAVbLNzWiy32sVNDqJRDiJK6klfDnAgu6PAGSHw==
 
 "@babel/core@7.12.9":
```
ankushdharkar, line 30: Since we haven't changed `package.json`, why do we have a new `yarn.lock`? Please look into it. This should ideally not have updated.

ankushdharkar (changes requested):

yarn.lock
akshay1502: I have raised the new commit removing the yarn.lock file.

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -48,8 +48,23 @@ const DragDropContextWrapper: FC<dragDropProps> = ({
   const ref = useRef<string[]>([]);
 
   useEffect(() => {
-    setTaskList(unAssignedTasks);
-    setMemberList(idleMembers);
+    const appendNewTasks = (
+      taskList: task[],
+      unAssignedTasks: task[]
+      ) :task[] => {
+        const oldTasksIds = taskList.map((Task: task) => Task.id);
```
harshith-venkatesh, line 55: Have we defined the interface in smallcase?

harshith-venkatesh: maybe taskItem or just item, Task looks bit different in my view

harshith-venkatesh: `oldTasksIds.includes(Task.id)`
Can't we directly do `!taskList.includes(taskItem/Task in this case?)

harshith-venkatesh (changes requested):
Good PR @akshay1502 , could you please look into why yarn.lock file is added here if we have not added any packages.
I have requested for few changes, please look into it

src/components/availability-panel/drag-drop-context/index.tsx
akshay1502: https://github.com/Real-Dev-Squad/website-status/blob/4367c70a1ef222bf9a5a1d163f16c1a902833452/src/interfaces/task.type.ts#L3
@harshith-venkatesh yes, the interface is defined in smallcase.

akshay1502: taskItem/Task is an object so can't check for them. So going with id here seems a better way to compare them.

harshith-venkatesh: https://bobbyhadz.com/blog/javascript-check-if-array-contains-object

harshith-venkatesh: Could you look into this and see if anything would be helpful to the above code?

harshith-venkatesh: I have rarely seen someone defining types in lowercase.
https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

Could this be looked into please
Either now or create a new ticket for the same

akshay1502: @harshith-venkatesh we need to filter out new tasks and old tasks here. With .some it will return us the boolean value but not the actual taksItem element. 
I am trying to filter the tasks whose id is not present in oldTasksIds array which becomes the new tasks, so as to append them with the oldTasks.

akshay1502: @harshith-venkatesh currently we have all the types in lowercase in `interfaces` folder. Changing them into caps, will require to change in all places where we have used it.

harshith-venkatesh: Got it , in that case I am ok with the changes and could you create a new issue and have it discussed with @ankushdharkar for the same please

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -48,8 +48,23 @@ const DragDropContextWrapper: FC<dragDropProps> = ({
   const ref = useRef<string[]>([]);
 
   useEffect(() => {
-    setTaskList(unAssignedTasks);
-    setMemberList(idleMembers);
+    const appendNewTasks = (
+      taskList: task[],
+      unAssignedTasks: task[]
+      ) :task[] => {
+        const oldTasksIds = taskList.map((taskItem: task) => taskItem.id);
```
harshith-venkatesh, line 55: NIT: oldTaskIDs?

harshith-venkatesh (changes requested):
Please remove yarn.lock
Rest looks fine

harshith-venkatesh (approved):

Pavangbhat (approved):

Issue #234, opened 2021-12-28
Handle New task workflow changes(Status site)


Pavangbhat:
**Details :-**  https://github.com/Real-Dev-Squad/website-backend/issues/460

**Tasks :-**

- [x] On the availability panel render only tasks that are available, not unAssigned(handle it before migration such that both AVAILABLE and unAssigned works)
- [x] When the task is assigned to a member from the availability panel make the status of the task as ASSIGNED and startedOn should be computed and endsOn should be 14 days by default
- [x]  **active → IN_PROGRESS, unAssigned → AVAILABLE, completed → VERIFIED** handle both scenarios such that **active**  behaves the same as **IN_PROGRESS** (Same for other changes)

PR #240, opened 2022-01-06
Update/task workflow 

Pavangbhat:
Closes: #234


src/helperFunctions/beautifyTaskStatus.ts
```diff
@@ -0,0 +1,19 @@
+import task from '../interfaces/task.type';
+
+const beautifyTaskStatus = (tasks: Array<task>) => {
+  const taskList: Array<task> = [];
+  tasks.forEach((taskItem: task, taskIndex) => {
+    taskList.push(taskItem);
+    const splitedStringArray = taskItem.status.split('_');
+    splitedStringArray.forEach((subString, index) => {
+      splitedStringArray[index] = subString.toLowerCase();
```
akshay1502, line 9: Better to make status as lowercase first and then use split.

src/helperFunctions/beautifyTaskStatus.ts
```diff
@@ -0,0 +1,19 @@
...
+    });
+    splitedStringArray.forEach((string, index) => {
+      splitedStringArray[index] = string.charAt(0).toUpperCase() + string.slice(1);
+    });
+    taskList[taskIndex].status = splitedStringArray.join(' ');
```
harshith-venkatesh, line 14: I feel like it can be handled in a single line provided the condition is how it is defined

```
const string = "aaaaa_masasdaas"

const convertData = (input) => input.toLowerCase().split('_').map(item => item.charAt(0).toUpperCase() + item.slice(1)).join(' ');

convertData(string)
console.log(convertData(string))
```

I was just trying it out , let me know your view on how readable or difficulty faced in reading.
But something to look into to make it more concise

src/pages/availability-panel.tsx
```diff
@@ -21,8 +22,9 @@ const AvailabilityPanel: FC = () => {
         const { requestPromise } = fetch({ url });
         const fetchPromise = await requestPromise;
         const { tasks } = fetchPromise.data;
-        const unassigned = tasks.filter(
-          (item: task) => item.status.toLowerCase() === 'unassigned' && item.type === 'feature',
+        const unassigned = updateTasksStatus(tasks).filter(
+          (item: task) => (
+            (item.status === 'AVAILABLE') && (item.type === 'feature')),
```
harshith-venkatesh, line 27: Could we not create a constant for `feature` and utilise the const `AVAILABLE` here?

src/helperFunctions/updateTasksStatus.ts
```diff
@@ -0,0 +1,23 @@
...
+    if ((taskList[index].status === 'active') || (taskList[index].status === 'pending')) {
+      taskList[index].status = 'IN_PROGRESS';
+    } else if (taskList[index].status === 'assigned') {
+      taskList[index].status = 'ASSIGNED';
+    } else if (taskList[index].status === 'unassigned') {
+      taskList[index].status = 'AVAILABLE';
+    } else if (taskList[index].status === 'completed') {
+      taskList[index].status = 'COMPLETED';
+    } else if (taskList[index].status === 'blocked') {
+      taskList[index].status = 'BLOCKED';
```
harshith-venkatesh, line 17: We can utilize the available constants defined here and probably convert that to lower case to handle the if condition here

harshith-venkatesh (changes requested):
Good work pavan, I have added few commits, do let me know your views and we can discuss if necessary

src/helperFunctions/beautifyTaskStatus.ts
harshith-venkatesh: I feel like it can be handled in a single line provided the condition is how it is defined

```
const string = "pawan_is_a_meMBER_oF_RDS"

const convertData = (input) => input.toLowerCase().split('_').map(item => item.charAt(0).toUpperCase() + item.slice(1)).join(' ');

convertData(string)
console.log(convertData(string))
```

I was just trying it out , let me know your view on how readable or difficulty faced in reading.
But something to look into to make it more concise

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -87,8 +87,13 @@ const DragDropcontext: FC<dragDropProps> = ({
         const assignee = result.combine.droppableId === 'tasks'
           ? result.draggableId
           : result.combine.draggableId;
+        const dateObject:any = new Date();
```
whyDontI, line 90: Let's try to drop the `any` wherever possible. Also, every class or interface can be used as a type in TypeScript.

So this line can be written as,
`const dateObject = new Date();`

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -87,8 +87,13 @@ const DragDropcontext: FC<dragDropProps> = ({
...
+        const startedOn:any = dateObject / 1000;
```
whyDontI, line 91: - I think it'll be a good idea to have a variable for `1000`, it'll be helpful to understand why are we dividing the `dateObject` by `1000`.

- I think we need to get epoch from the date before performing the division. e.g. `dateObject.getTime() / 1000`

I am getting this error in `tsserver - eslint`
> The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.

src/components/availability-panel/drag-drop-context/index.tsx
```diff
@@ -87,8 +87,13 @@ const DragDropcontext: FC<dragDropProps> = ({
...
+        const endsOn:number = startedOn + (14 * 86400);
```
whyDontI, line 92: Let's put `86400` and `14` in a file-level global variable

whyDontI (changes requested):

src/helperFunctions/beautifyTaskStatus.ts
Pavangbhat: This is really easy to understand, will update it 
Thank you

src/pages/availability-panel.tsx
Pavangbhat: We should use here 👍

src/helperFunctions/updateTasksStatus.ts
Pavangbhat: Yes sure will be done

src/components/availability-panel/drag-drop-context/index.tsx
Pavangbhat: Got it will update it

src/components/availability-panel/drag-drop-context/index.tsx
Pavangbhat: Got it
> I will have a separate variable for 1000
> `I think we need to get epoch from the date before performing the division. e.g. dateObject.getTime() / 1000` :- I also thought so but then tried it and it was working, but anyway its good to avoid confusions will update it

src/components/availability-panel/drag-drop-context/index.tsx
Pavangbhat: Got it

src/helperFunctions/beautifyTaskStatus.ts
Pavangbhat: Thank you 
Will update it 👍

whyDontI (changes requested):
Just one little change, Everything else looks perfect 🥳

src/components/tasks/card/index.tsx
```diff
@@ -22,6 +16,8 @@ const Card: FC<Props> = ({
   shouldEdit = false,
   onContentChange = () => undefined,
 }) => {
+  const statusRedList = ['Blocked'];
+  const statusNotOverDueList = ['Completed', 'Verified', 'Available'];
```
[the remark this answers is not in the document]
Pavangbhat: Done

Pavangbhat:
continued with #248

PR #277, opened 2022-07-26
added datepicker in task card

### LINKED ISSUE #269: Create Date picker instead writing dates manually
In Status site, Instead of writing dates by formate wise, it should have a date picker, So that user can write dates by date picker.

ETA: 15 july

DeepanshuSarswat:
Closes #269 
Discription: created date picker instead writing dates manually.

https://user-images.githubusercontent.com/78958441/181081624-a59296f5-38df-4de9-a422-86d6278f5ccf.mp4


src/components/tasks/card/index.tsx
```diff
@@ -49,9 +54,22 @@ const Card: FC<Props> = ({
...
+  
+    let inputDate = event.target.value;
+    let Date_to_String = new Date(inputDate);
+    
+    Date_to_String.setHours(23, 59, 59, 999);
+
+    let StringToDate = (JSON.stringify(Date_to_String)).slice(1,11)
+
+let [year, month, day] = StringToDate.split('-');
+let resultDate = [month, day, year].join('/');
```
rohan09-raj, line 67: Please use const wherever we can

src/components/tasks/card/index.tsx
```diff
@@ -66,6 +84,8 @@ const Card: FC<Props> = ({
   if (isTaskOverdue()) {
     cardClassNames.push(classNames.overdueTask);
   }
+
```
rohan09-raj, line 87: Can we please remove this extra line ?

src/components/tasks/card/index.tsx
```diff
@@ -36,6 +36,11 @@ const Card: FC<Props> = ({
 
   const cardClassNames = [classNames.card];
 
+  let changeToDate = new Date(localEndsOn);
+  let dates = (JSON.stringify(changeToDate)).slice(1,11)
```
rohan09-raj, line 40: Can we please use `const` here if we can ?

src/components/tasks/card/index.tsx
```diff
@@ -106,15 +126,19 @@ const Card: FC<Props> = ({
...
-            tabIndex={0}
-          >
-            {fromNowEndsOn}
-          </span>
+          {shouldEdit ? (
+            <input type="date" onChange={(e) => handleChange(e, "endsOn")}
+            onKeyPress={(e) => handleChange(e, "endsOn")}
+            value ={dateTime}  />
+          ) : (
+            <span
```
rohan09-raj, line 134: Can we please format the code using Prettier or some other code formatter ?

src/components/tasks/card/index.tsx
```diff
@@ -36,6 +36,11 @@ const Card: FC<Props> = ({
```
rohan09-raj, line 36: Good work @DeepanshuSarswat 
There is one fix that is needed to be done
Rest everything is working fine
When I select 24-08-2022 and save it, the date saved is 23-08-2022 means every time I save the date, the actual date saved is a day before 

image

src/components/tasks/card/index.tsx
```diff
@@ -21,40 +26,61 @@ const Card: FC<Props> = ({
...
+    : "#f83535";
+  const iconHeight = "25px";
+  const iconWidth = "25px";
 
   const cardClassNames = [classNames.card];
 
+  const changeToDate = new Date(localEndsOn);
+  const dates = JSON.stringify(changeToDate).slice(1, 11);
+
+  const [dateTime, setdateTime] = useState(dates);
```
rohan09-raj, line 50: `D` can be in capitals here

src/components/tasks/card/index.tsx
```diff
@@ -21,40 +26,61 @@ const Card: FC<Props> = ({
...
-    return tmp.textContent || tmp.innerText || '';
+    return tmp.textContent || tmp.innerText || "";
   }
 
   function handleChange(event: any, changedProperty: keyof typeof cardDetails) {
-    if (event.key === 'Enter') {
+    setdateTime(event.target.value);
+
+    const inputDate = event.target.value;
+    const DateToString = new Date(inputDate);
```
rohan09-raj, line 68: Typo `dateToString`

src/components/tasks/card/index.tsx
DeepanshuSarswat: @rohan09-raj , I have resolve the problem.
Please have a look.

src/components/tasks/card/index.tsx
DeepanshuSarswat: I've done these task.
And I have updated PR again.
Please, have a look

src/components/tasks/card/index.tsx
```diff
@@ -1,10 +1,15 @@
...
+import task from "@/interfaces/task.type";
+import {
+  AVAILABLE,
+  BLOCKED,
+  COMPLETED,
+  VERIFIED,
+} from "@/components/constants/beautified-task-status";
 
-const moment = require('moment');
+const moment = require("moment");
```
rohan09-raj, line 12: Can we please use single quotes as they were before throughout the changed files ? You can change the settings of your local prettier in the VSCode settings

DeepanshuSarswat: Done!

src/components/tasks/card/index.tsx
```diff
@@ -100,52 +112,55 @@ const Card: FC<Props> = ({
...
-            role="button"
-            tabIndex={0}
-          >
-            {fromNowEndsOn}
-          </span>
+          {shouldEdit ? (
+            <input
+              type='date'
+              onChange={(e) => handleChange(e, 'endsOn')}
+              onKeyPress={(e) => handleChange(e, 'endsOn')}
```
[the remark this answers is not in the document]
DeepanshuSarswat: onChange and onKeyPress is not same.we need both

ankushdharkar (changes requested):
Too many prettier changes. Could we just have the changes relevant to your feature please?

src/components/tasks/card/index.tsx
```diff
@@ -35,6 +35,8 @@ const Card: FC<Props> = ({
   const iconWidth = '25px';
 
   const cardClassNames = [classNames.card];
+  const dates = JSON.stringify(localEndsOn).slice(1, 11);
```
ankushdharkar, line 38: Will this result in runtime exception if `localEndsOn` is nullish or an unexpected value?

ankushdharkar (changes requested):

src/components/tasks/card/index.tsx
DeepanshuSarswat: Done!

Maheima (approved):
Looks good to me

src/components/tasks/card/index.tsx
```diff
@@ -106,15 +108,22 @@ const Card: FC<Props> = ({
...
-          <span
-            className={classNames.cardStrongFont}
-            contentEditable={shouldEdit}
-            onKeyPress={(e) => handleChange(e, 'endsOn')}
-            role="button"
-            tabIndex={0}
-          >
-            {fromNowEndsOn}
-          </span>
+          {shouldEdit ? (
```
Maheima, line 111: When does shouldEdit become true?

DeepanshuSarswat: when user is Authorized and int url edit=true
both will true then shouldEdit bexomes true

Maheima: is any user allowed to do it or just super user/admin

src/components/tasks/card/index.tsx
```diff
@@ -35,6 +35,8 @@ const Card: FC<Props> = ({
   const iconWidth = '25px';
 
   const cardClassNames = [classNames.card];
+  const dates = localEndsOn == undefined? " " : JSON.stringify(localEndsOn).slice(1, 11);
```
ankushdharkar, line 38: How many dates are here?

src/components/tasks/card/index.tsx
```diff
@@ -106,15 +108,22 @@ const Card: FC<Props> = ({
...
+            />
+          ) : (
+            <span
+              className={classNames.cardStrongFont}
+              role='button'
+              tabIndex={0}
+            >
+              {fromNowEndsOn}
+            </span>
+          )}
```
ankushdharkar, line 126: Nit: For simplicity, you can actually move this if-else logic into a function `renderDate(isEditing)`. It will make it easier to read this current component and abstract the logic of showing picker/span into it's own single responsibility function.

src/components/tasks/card/index.tsx
ankushdharkar: Can this be written a bit more cleanly like:

```js
 const x = !!localEndsOn ? JSON.stringify(localEndsOn).slice(1, 11) : '';
```

First the part of what should happen, not the empty case first.

ankushdharkar: I also don't understand what this JSON stringify is doing? Why do we need it?

ankushdharkar (changes requested):

src/components/tasks/card/index.tsx
```diff
@@ -35,6 +35,8 @@ const Card: FC<Props> = ({
   const iconWidth = '25px';
 
   const cardClassNames = [classNames.card];
+  const dates = localEndsOn == undefined? " " : JSON.stringify(localEndsOn).slice(1, 11);
+  const [dateTime, setDateTime] = useState(dates);
```
ankushdharkar, line 39: Why are multiple dates used for a single dateTime?

src/components/tasks/card/index.tsx
DeepanshuSarswat: typeof localEndsOn is object and its gives us "Thu Aug 11 2022 15:37:00 GMT+0530 (India Standard Time)"
We want to extract only date from it.
Example '2022-08-11'

 we will use slice so that we can get selected part of the string but slice can only call on the string or the array.
since localEndOn is Object, so first we will change it into string with he help of JSON.stringify().

Now, We get string and we can use slice method.

When we call slice method on undefined or object then we get Type Error.

This is the main reason, to use JSON.stringify() in the code

DeepanshuSarswat: There are many dates.
because, Every user will have deadline.

DeepanshuSarswat: Yea!,
 const x = !!localEndsOn ? JSON.stringify(localEndsOn).slice(1, 11) : '';
 This is much simple and  working on both "null" and "undefined" .
 Thanks!

src/components/tasks/card/index.tsx
DeepanshuSarswat: All these data coming from "content" props.  we iterate this, then get these multiple values.
And this is fine because we are putting result of Array iteration.

src/components/tasks/card/index.tsx
DeepanshuSarswat: Done!

src/components/tasks/card/index.tsx
ankushdharkar: "Nit"

src/components/tasks/card/index.tsx
```diff
@@ -105,16 +130,8 @@ const Card: FC<Props> = ({
...
-            className={classNames.cardStrongFont}
-            contentEditable={shouldEdit}
-            onKeyPress={(e) => handleChange(e, 'endsOn')}
-            role="button"
-            tabIndex={0}
-          >
-            {fromNowEndsOn}
-          </span>
+          <span className={classNames.cardSpecialFont}>Due Date</span>      
+         {renderDate(fromNowEndsOn,shouldEdit)}
```
harshith-venkatesh, line 134: Wouldn't it be better to create a component passing the props?

src/components/tasks/card/index.tsx
```diff
@@ -66,6 +68,29 @@ const Card: FC<Props> = ({
   if (isTaskOverdue()) {
     cardClassNames.push(classNames.overdueTask);
   }
+
+  function renderDate(fromNowEndsOn: string, shouldEdit: boolean){
+    if(shouldEdit){
```
harshith-venkatesh, line 73: Wouldn't it be better to create a ternary operator?
It seems like it is either this or that right?

harshith-venkatesh (changes requested):

src/components/tasks/card/index.tsx
DeepanshuSarswat: @harshith-venkatesh , this is only 16 line code. 
Do you think, we should create component for this?

harshith-venkatesh: Ok I agree with you
Couldn't we have gone with ternary condition here instead of function? As I observe based on shouldEdit we are showing snippet right?

src/components/tasks/card/index.tsx
DeepanshuSarswat: I did,
Isn't this method cleaner and simpler than that?
image

harshith-venkatesh: Doesn't this snippet does the entire work of the function created?

harshith-venkatesh: And if we wouldn't be utilising anywhere else, then having this snippet is better than a separate function call and returning the same.

If we had multiple functions, then it would be good to have a separate function like one created if the scope of card is limited then yes this sound right

src/components/tasks/card/index.tsx
```diff
@@ -66,6 +68,29 @@ const Card: FC<Props> = ({
...
+      )
+    } 
+    return(  
+      <span
+          className={classNames.cardStrongFont}
+          role='button'
+          tabIndex={0}
+        >
+          {fromNowEndsOn}
+        </span>
```
ankushdharkar, line 90: Indentation is not right @DeepanshuSarswat

ankushdharkar (changes requested):

src/components/tasks/card/index.tsx
```diff
@@ -34,7 +34,17 @@ const Card: FC<Props> = ({
   const iconHeight = '25px';
   const iconWidth = '25px';
 
-  const cardClassNames = [classNames.card];
+function FormattedDates(localEndsOn: Date){
```
ankushdharkar, line 37: Why is the function name not camelCased?

ankushdharkar (changes requested):

src/components/tasks/card/index.tsx
```diff
@@ -34,7 +34,17 @@ const Card: FC<Props> = ({
...
+  const year = localEndsOn.getFullYear()
+  const month = localEndsOn.getMonth() + 1;
+  const day = localEndsOn.getDate();
+  const finalDate = [year, month, day].join('-');
+  return finalDate;
+}
+
+  const cardClassNames = [classNames.card];
```
ankushdharkar, line 45: Where is this used?

src/components/tasks/card/index.tsx
```diff
@@ -34,7 +34,16 @@ const Card: FC<Props> = ({
...
+function formattedDates(localEndsOn: Date){
+  const year = localEndsOn.getFullYear()
+  const month = localEndsOn.getMonth() + 1;
+  const day = localEndsOn.getDate();
+  const finalDate = [year, month, day].join('-');
+  return finalDate;
+}
+
+  const dates = !!localEndsOn ? formattedDates(localEndsOn) : '';
+  const [dateTimes, setDateTimes] = useState(dates);
```
heyrandhir, line 46: Few Suggestions for this Function : 

1. move this function in a util.
2. function name can be improved to something like getDateInStringFormat
3. parameter to the function can be named date instead of localEndsOn
4. finalDate can be named better something like strDate.

src/components/tasks/card/index.tsx
```diff
@@ -34,7 +34,16 @@ const Card: FC<Props> = ({
...
-  const cardClassNames = [classNames.card];
+function formattedDates(localEndsOn: Date){
+  const year = localEndsOn.getFullYear()
+  const month = localEndsOn.getMonth() + 1;
+  const day = localEndsOn.getDate();
+  const finalDate = [year, month, day].join('-');
+  return finalDate;
+}
+
+  const dates = !!localEndsOn ? formattedDates(localEndsOn) : '';
```
heyrandhir, line 45: name of the variable can be changed to strDate to give a better sense and the type can be defined as string.

ankushdharkar (approved):

ankushdharkar:
Thanks for sticking around through all the iterations! Looks good to be 🚢 !

PR #283, opened 2022-08-18
Render date

akshay1502:
Using date formating in 
- tasks route
- openPr
- stalePr
- mine

in tasks,mine we have dates in number format
while for openPr, stalePr we have dates in string format
Handling both number and string format of dates.
## before
datebeforerds

## after
dateafterrds


akshay1502:
https://github.com/redwoodjs/redwood/issues/5104#issuecomment-1094347397
https://stackoverflow.com/a/71828113/13964048
Added resultions due to following error.
Screenshot (200)
l

src/utils/renderDate.ts
```diff
@@ -0,0 +1,10 @@
+const moment = require('moment');
+// dates are in number and string format.
+function dateFromNow(date: string | number): string {
+  if (date === null) return 'Invalid date';
```
ivinayakg, line 4: instead of string can you return undefined or null I guess that would be easy to handle error or edge cases

package.json
```diff
@@ -51,9 +51,8 @@
     "@storybook/react": "^6.4.9",
     "@types/next": "^9.0.0",
     "@types/node": "^17.0.5",
-    "@types/react": "^17.0.38",
-    "@types/react-beautiful-dnd": "^13.1.2",
     "@types/react-dom": "^17.0.11",
+    "@types/react-beautiful-dnd": "^13.1.2",
```
ivinayakg, line 55: don't push changes of package.json or yarn.lock into your branch

ivinayakg (changes requested):

src/utils/renderDate.ts
akshay1502: Rather we can have a better message to show. `Invalid date` shows up in the UI for available tasks as they don't have startedOn, endsOn. Either we can hide showing due date and startedOn OR we can show a better message over it.
What do you think @ivinayakg ?

Maheima:
LGTM

ivinayakg (approved):

ivinayakg:
LGTM

ankushdharkar:
Why was the `yarn.lock` deleted??

ivinayakg (changes requested):
@akshay1502 remove the yarn.lock file changes from the pr

akshay1502:
Now I know how to deal with yarn.lock.

ivinayakg (approved):

PR #289, opened 2022-08-21
Fix/dateformat

DeepanshuSarswat:
Fixed date format

### DIFF
--- a/src/helperFunctions/getDateInString.ts
+++ b/src/helperFunctions/getDateInString.ts
@@ -1,7 +1,7 @@
 const getDateInString = (date: Date) => {
     const year = date.getFullYear()
-    const month = ("0" + (date.getMonth() + 1)).slice(-2);
-    const day = ("0" + date.getDate()).slice(-2);
+    const month = date.getMonth() + 1;
+    const day = date.getDate();
     const strDate = [year, month, day].join('-');
     return strDate;
 };
--- a/src/helperFunctions/getDateInString.ts
+++ b/src/helperFunctions/getDateInString.ts
@@ -1,7 +1,7 @@
 const getDateInString = (date: Date) => {
     const year = date.getFullYear()
-    const month = date.getMonth() + 1;
-    const day = date.getDate();
+    const month = ("0" + (date.getMonth() + 1)).slice(-2);
+    const day = ("0" + date.getDate()).slice(-2);
     const strDate = [year, month, day].join('-');
     return strDate;
 };


src/helperFunctions/getDateInString.ts
```diff
@@ -1,7 +1,7 @@
 const getDateInString = (date: Date) => {
     const year = date.getFullYear()
-    const month = date.getMonth() + 1;
-    const day = date.getDate();
+    const month = ("0" + (date.getMonth() + 1)).slice(-2);
```
harshith-venkatesh, line 3: Would ternary operator be better suitable here?
Checking the length and adding 0 or not accordingly?
A new dev find it difficult to understand without comments or context

harshith-venkatesh (changes requested):

src/helperFunctions/getDateInString.ts
DeepanshuSarswat: Like this?
const a = new Date();
const month = a.getMonth > 9 ? a.getMonth : ("0" + (a.getMonth() + 1)).slice(-2);

ankushdharkar: @DeepanshuSarswat Please try to use code blocks in the comments. It helps with readability.

@harshith-venkatesh I actually really like the solution used here. A comment would help. I feel a ternary would be an overkill.

```js
const month = ("0" + (date.getMonth() + 1)).slice(-2);
```

harshith-venkatesh: True, after taking a look at the comment back, I felt the same as well.
However the comment do play an important role here

ankushdharkar: @harshith-venkatesh I feel at the end of the day, we are not writing code for new devs. We are writing code that is easy to read, proven correct (Most important) and reason about.

A new dev should take the time to look up what the language functions do and if they can guarantee that the code will always produce the correct results.

If we always cater to the new dev, we will be stuck writing verbose basic code.

ankushdharkar (approved):
Very smart and elegant fix! Thanks @DeepanshuSarswat

src/helperFunctions/getDateInString.ts
harshith-venkatesh: Can't disagree with your comment. That intention here was, I know the reason behind this fix.
But a person new to review or with no context should also be able to easily review and approve right.
I can understand, this statement could be easily deciphered, was just presenting my thoughts here🙂👍

harshith-venkatesh:
Good work, I was about to approve this PR as well, it got merged already 😅

PR #306, opened 2022-09-07
Refactored isUserAuthorised

akshay1502:
Refactored the isUserAuthorised function by moving to _app.tsx on top and sharing the state into the below pages.


harshith-venkatesh:
Could you check if CONTEXT API would be better approach here?

akshay1502:
@harshith-venkatesh going with context API, thanks for suggesting the approach. Since we might need isUserAuthorized functionality at many places deep down in the components.

harshith-venkatesh:
Yes @akshay1502 , you did come to a great conclusion and you are also right with the reason

Maheima (approved):
LGTM! Good work @akshay1502 !

src/pages/_app.tsx
```diff
@@ -1,20 +1,53 @@
...
+const SELF_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/users/self`;
+const { ERROR } = ToastTypes;
+export const isUserAuthorizedContext = createContext<Boolean>(false);
+
+const MyApp: FC<AppProps> = ({ Component, pageProps }) => {
+  const [isUserAuthorized, setIsUserAuthorized] = useState(false);
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        console.log('checking for superuser from index pages');
```
harshith-venkatesh, line 23: Remove the console log

src/pages/_app.tsx
```diff
@@ -1,20 +1,53 @@
...
+    };
+    fetchData();
+
+    return (() => {
+      setIsUserAuthorized(false);
+    });
+  }, []);
+  return (
+    <AppWrapperContext>
+      <isUserAuthorizedContext.Provider value={isUserAuthorized}>
```
harshith-venkatesh, line 44: This is not a good architectural way of designing context and will also lead to bulk MyApp page
Could you please ensure to create a separate file to handle this situation

harshith-venkatesh (changes requested):

src/pages/_app.tsx
akshay1502: 😅done.

src/pages/_app.tsx
akshay1502: done.

src/components/tasks/card/index.tsx
```diff
@@ -1,14 +1,13 @@
-import { FC, useState, useEffect } from 'react';
+import { FC, useState, useEffect, useContext } from 'react';
 import Image from 'next/image';
 import classNames from '@/components/tasks/card/card.module.scss';
 import task from '@/interfaces/task.type';
 import { AVAILABLE, BLOCKED, COMPLETED, VERIFIED } from '@/components/constants/beautified-task-status';
 import getDateInString from '@/helperFunctions/getDateInString';
```
Maheima, line 6: can we sort the imports here somewhat like in the format where
the library imports are first
then components
then CSS followed by assets

src/pages/_app.tsx
```diff
@@ -1,20 +1,53 @@
...
-  <AppWrapperContext>
-    <ToastContainer />
-    <Component {...pageProps} />
-  </AppWrapperContext>
-);
+const SELF_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/users/self`;
+const { ERROR } = ToastTypes;
+export const isUserAuthorizedContext = createContext<Boolean>(false);
+
+const MyApp: FC<AppProps> = ({ Component, pageProps }) => {
```
Maheima, line 13: Can we create another file to create this context. I don't think we should put this much code in _app.tsx ? Check AppWrapperContext for reference

Maheima (changes requested):

src/pages/_app.tsx
akshay8902: I am using isUserAuthorized file for the context, and using it here importing it.

akshay1502: @Maheima i am using isUserAuthorized file for context and using it by importing here.

src/components/tasks/card/index.tsx
akshay1502: Sure, will sort them.

src/pages/_app.tsx
Maheima: Awesome!

src/context/isUserAuthorized.tsx
```diff
@@ -0,0 +1,43 @@
+import { useEffect, useState, createContext, FC, Children, ReactNode } from "react";
```
Maheima, line 1: this Children is not used anywhere

akshay1502: @Maheima we are using it on line no.38

Maheima (approved):
LGTM

Maheima:
@harshith-venkatesh awaiting your review of this PR. I am merging it for now as it resolves a critical issue

Maheima:
> @harshith-venkatesh awaiting your review of this PR. I am merging it for now as it resolves a critical issue

https://github.com/Real-Dev-Squad/website-status/issues/333

src/context/isUserAuthorized.tsx
```diff
@@ -0,0 +1,43 @@
...
+import fetch from '@/helperFunctions/fetch';
+import { toast, ToastTypes } from '@/helperFunctions/toast';
+
+const SELF_URL = `${process.env.NEXT_PUBLIC_BASE_URL}/users/self`;
+const { ERROR } = ToastTypes;
+export const isUserAuthorizedContext = createContext<Boolean>(false);
```
harshith-venkatesh, line 7: Few naming convention that is followed which is good to have in the project.
1. Context are name with Capital case IsUserAuthorizedContext , and we could go ahead with IsUserAuthorizedContextProvider which would be IsUserAuthorizedContext.Provider

src/context/isUserAuthorized.tsx
```diff
@@ -0,0 +1,43 @@
...
+      }
+    };
+    fetchData();
+
+    return (() => {
+      setIsUserAuthorized(false);
+    });
+  }, []);
+  return(
+    <isUserAuthorizedContext.Provider value={isUserAuthorized}>
```
harshith-venkatesh, line 37: We should have also passed setIsUserAuthorized along side to ensure that in future when we use logout button it would be useful and there won't be any refactor of code as such

PR #308, opened 2022-09-10
fix: handled keydown press for enter and space in accordian component

prakashchoudhary07:
Closes https://github.com/Real-Dev-Squad/website-status/issues/309

 Weird behavior when alt key is pressed to go to edit mode by the superuser


harshith-venkatesh (approved):

src/components/constants/key.ts
```diff
@@ -1 +1,3 @@
-export const ALT_KEY = 'Alt';
```
Maheima, line 1: why does it show a change in this line even when they are identical?

ankushdharkar (approved):

Issue #304, opened 2022-09-05
Infinite fetching of images on idle members page.


ivinayakg:
After the fetch is failed for the image for the members on the idle-members page, the browser goes into infinite re-fetching for the photos.
Also if there is an error in image fetching the default `DUMMY_PROFILE` is not visible.

PR #312, opened 2022-09-12
Infinite Image fetch fixed

ivinayakg:
RFC - #304

## What was happening Before
Whenever there is an error in getting the image of a user on the `idle-members` page it used to re-fetch infinitely as the error handling was done in a `non-React way` and therefore was not working.

## What I did do and How is it working?
I just simply added a state to the component so that If It finds any error, it will re-render with a fetching fallback image for error handling.

### DIFF
--- a/src/components/idleMembers/card/index.tsx
+++ b/src/components/idleMembers/card/index.tsx
@@ -10,7 +10,7 @@ type Props = {
 }
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
-  const [isError, setIsError] = useState(false);
+  const [isImageAvailable, setIsImageAvailable] = useState(false);
 
   const assigneeProfilePic = (name: string) => `${IMAGE_URL}/${name}/img.png`;
   const getMemberDetails = (name: string) => {
@@ -18,7 +18,7 @@ const Card: FC<Props> = ({ idleMemberUserName }) => {
     if (newWindow) newWindow.opener = null;
   };
   const assigneeImageOnError = (e: SyntheticEvent<HTMLImageElement>) => {
-    setIsError(true);
+    setIsImageAvailable(true);
   };
   return (
     <div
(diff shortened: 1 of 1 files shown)


src/components/idleMembers/card/index.tsx
```diff
@@ -10,13 +10,15 @@ type Props = {
...
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
+  const [isError, setIsError] = useState(false);
+
   const assigneeProfilePic = (name: string) => `${IMAGE_URL}/${name}/img.png`;
   const getMemberDetails = (name: string) => {
     const newWindow = window.open(`https://members.realdevsquad.com/${name}`, '_blank', ' noopener ,norefferrer');
     if (newWindow) newWindow.opener = null;
   };
   const assigneeImageOnError = (e: SyntheticEvent<HTMLImageElement>) => {
```
harshith-venkatesh, line 20: Where is this function being called?

src/components/idleMembers/card/index.tsx
```diff
@@ -10,13 +10,15 @@ type Props = {
 }
 
 const Card: FC<Props> = ({ idleMemberUserName }) => {
+  const [isError, setIsError] = useState(false);
```
harshith-venkatesh, line 13: NIT: Could we think of a better variable aligning with the Image and source?
I see that we are utilizing this for Image Error scenario, so something on that line :)

harshith-venkatesh (approved):
I am approving this PR as it is meeting the requirements.
I had a question and an NIT
Please look into that 
Great PR and hope you had good time debugging and documenting

src/components/idleMembers/card/index.tsx
vinayak-trivedi, line 13: we can use `isImageAvailable, setIsImageAvailable` here, instead of `error, isError`

ivinayakg: yeah sure will do that

harshith-venkatesh:
Please resolve the conversation or reply back once completed

src/components/idleMembers/card/index.tsx
ivinayakg: On the image tag in the jsx

harshith-venkatesh (approved):

PR #369, opened 2022-12-07
refactoring and filtering options in edit mode

Muralidhar22:
## Change
- Refactoring the level and item keys
- Filtering out tag options if it's already present on the task (In edit mode of tasks page on status site)
Screenshot (16)

Fig: System design is not present in the below options as it was already present on the task.


prakashchoudhary07 (approved):
LGTM

vinayak-trivedi (approved):
LGTM

PR #378, opened 2022-12-21
Return sorted users on the basis of how long they have been idle for

Maheima:
This PR fixes #344 

Changes made:
1. Sorting option changed from Alphabetically to the smallest date
2. Changed the folder and file names from idle-members, idleMembers -> idle-users, idleUsers respectively
3. Idle user API from `https://api.realdevsquad.com/members/idle` -> `https://api.realdevsquad.com/users/status?state=IDLE`
4. Added a helper function to get idle Since Text
5. Added tests to check the functionality

Is dev tested : Yes

image
Follow up issue of [showing user profile on click](https://github.com/Real-Dev-Squad/website-status/issues/376)

PR #380, opened 2022-12-23
Show user profile on click

Maheima:
This needs to be merged after https://github.com/Real-Dev-Squad/website-status/pull/378

Changes made:

1. The test case for link written
2. Fixed the date bug in the PR sent right before this one
3. Added members URL

For members, the profile is going to be visible to public but for users, it is going to be displayed only to super user/Admin. This is implemented in members site already.


heyrandhir (dismissed):
lgtm

src/components/idleUsers/section/index.tsx
```diff
@@ -1,21 +1,15 @@
...
-}
-
-function renderCards(content: Array<string>) {
-  return content.map((idleMember) => (
-    <Card idleMemberUserName={idleMember} key={idleMember} />
+function renderCards(content: idleUser[]) {
+  return content.map((idleUser) => (
+    <Card user={idleUser} key={idleUser.id} />
   ));
 }
```
yesyash, line 10: Should this be not wrapped inside a fragment?

src/components/idleUsers/section/index.tsx
```diff
@@ -1,21 +1,15 @@
...
 
-const Section: FC<Props> = ({
+const Section: FC<idleUserArray> = ({
```
yesyash, line 12: `idleUserArray` might not exit after PR #378 will be merged. 
[Link](https://github.com/Real-Dev-Squad/website-status/pull/378/files#diff-46824615da47fe1362ea6861d10da582e83535582d85047f1a814b6da4f3037b)

src/helperFunctions/convertDatetoYMD.ts
```diff
@@ -0,0 +1,30 @@
...
+  years = year - yy;
+  if (month * 100 + day < mm * 100 + dd) {
+    years = years - 1;
+    months = months + 12;
+  }
+  // days
+  const days = Math.floor(
+    (today.getTime() - new Date(yy + years, mm + months - 1, dd).getTime()) /
+      (24 * 60 * 60 * 1000)
+  );
```
yesyash, line 27: #### Optional: Can we use a library instead to perform this logic?

src/interfaces/idleUser.type.ts
```diff
@@ -0,0 +1,22 @@
+export type idleUserArray = {
```
yesyash, line 1: Name changed in PR #378

yesyash (dismissed):

src/helperFunctions/convertDatetoYMD.ts
heyrandhir: This logic will go away now as we will now show the user idle duration in just days 😅

src/components/idleUsers/section/index.tsx
Maheima: I don't think so. Wrapping is necessary when you have multiple tags

heyrandhir (approved):
LGTM. 🙂👍🏻

ivinayakg (dismissed):
Looks good to merge 👍

src/components/idleUsers/card/card.module.scss
```diff
@@ -8,16 +11,23 @@
   flex-wrap: wrap;
   margin: 2rem;
   padding: 2rem;
-  cursor: pointer;
```
RitikJaiswal75, line 11: was this intentional?
the cursor pointer gives the feel of a clickable object.

src/components/idleUsers/card/card.module.scss
```diff
@@ -8,16 +11,23 @@
...
   align-items: center;
 
   img {
     border-radius: 50%;
   }
 
   .name {
+    cursor: pointer;
```
RitikJaiswal75, line 21: What if I am hovering on the image, it should be a cursor pointer rather than just keeping it on the name.

src/components/constants/date.js
```diff
@@ -1,5 +1,5 @@
 export const THOUSAND_MILLI_SECONDS = 1000;
```
RitikJaiswal75, line 1: do we really need this? isn't it a predefined quantity? 
instead of such big constant name it will be easy to write 1000

src/components/idleUsers/card/card.module.scss
Maheima: Yes

src/components/idleUsers/card/card.module.scss
Maheima: I have made only name clickable not the whole card

src/components/constants/date.js
Maheima: It isn't a part of my PR.

RitikJaiswal75: I guess an AO could make little changes and put it up here.

Maheima: But I feel ritik, it is telling us about an entity in the formula we are applying. So, it is good to put in a constant.

yesyash (approved):

harshith-venkatesh (approved):

src/components/constants/date.js
harshith-venkatesh: If this is being used in multiple places and provides info to the devs on the importance of it we can keep it.
Ideally we don't need Thousand Milli seconds or fourteen days as it doesn't add much value.

harshith-venkatesh: 3,4,5 makes much sense and it is needed as well

Issue #375, opened 2022-12-18
User tasks not being shown on "/mine" route


yesyash:
### Description
When a user who has already been assigned tasks visits the `/mine` route there are no tasks being displayed even if we get the required data from the `/self` API. 

#### Image For reference: 
![image](https://user-images.githubusercontent.com/56453897/208294623-3f2ef7df-6d7b-4280-9a10-a147e868a8a5.png)

HeyNitin:
Hi @Maheima , can you please assign it to me

kshitij430:
Hi, Please assign this task to me.

kshitij430:
PR [#386](https://github.com/Real-Dev-Squad/website-status/pull/386), fixes the issue. Please review.

sonai99:
Hey @Maheima 
You can assign this to me.

yesyash:
@Maheima is this issue resolved by pr #386?
Can we close this?

sonai99:
@yesyash I see this is the latest fix https://github.com/Real-Dev-Squad/website-status/pull/414 for this issue
This is also not merged and has 2 Unresolved conversations. Do let me know if you still want me to work on this issue
@Maheima

Maheima:
Hey @sonai99 ,
That is the PR I gave you in reference to assigning you this ticket. Can you please continue the work on that which will involve you verifying what has been done, checking if it works fine, replying to any unresolved conversations and getting it reviewed and merged

PR #386, opened 2023-01-08
fixed mine route issue

kshitij430:
This PR fixes issue https://github.com/Real-Dev-Squad/website-status/issues/375

There were two issues .
1) The Mine component was re evaluating in a loop as the useEffect hook had "response" as  dependency.
2)  The if condition to check if there is a response with length had a not condition which did the opposite. Ex - for 0 results found !0 would evaluate to 1 and !1 would evaluate to a falsy value due to which the setTasks state updating function was not getting executed.

Fixes I have done -
1) Removed response as a dependency as we only want to call the API once when the component is mounted.
2) Removed the condition to check response results as these checks are already done in the JSX return. Ex-> display no results found if length is 0 and otherwise.

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,4 +1,4 @@
-NEXT_PUBLIC_BASE_URL='https://api.realdevsquad.com'
+NEXT_PUBLIC_BASE_URL='https://staging-api.realdevsquad.com'
 NEXT_PUBLIC_GITHUB_IMAGE_URL='https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members'
 NEXT_PUBLIC_API_MOCKING='ON'
 
--- a/.env.development
+++ b/.env.development
@@ -1,4 +1,4 @@
-NEXT_PUBLIC_BASE_URL='https://staging-api.realdevsquad.com'
+NEXT_PUBLIC_BASE_URL='https://api.realdevsquad.com'
 NEXT_PUBLIC_GITHUB_IMAGE_URL='https://raw.githubusercontent.com/Real-Dev-Squad/website-static/main/members'
 NEXT_PUBLIC_API_MOCKING='ON'
 
--- a/src/pages/mine.tsx
+++ b/src/pages/mine.tsx
@@ -32,7 +32,7 @@ const Mine: FC = () => {
   const { isLoggedIn, isLoading: isAuthenticating } = useAuthenticated();
 
   useEffect(() => {
-    if (isLoggedIn && Object.keys(response).length) {
+    if (isLoggedIn) {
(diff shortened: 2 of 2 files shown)


kshitij430:
> Please add description to the PR related to why it was happening and how it was fixed

Already Done. Can you please review it?

Maheima (changes requested):

.env.development
```diff
@@ -1,4 +1,4 @@
-NEXT_PUBLIC_BASE_URL='https://staging-api.realdevsquad.com'
+NEXT_PUBLIC_BASE_URL='https://api.realdevsquad.com'
```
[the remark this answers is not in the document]
kshitij430: Done. :(

ivinayakg (dismissed):
perfect

vvaibhavdesai (dismissed):

heyrandhir (dismissed):

src/pages/mine.tsx
```diff
@@ -32,11 +32,11 @@ const Mine: FC = () => {
   const { isLoggedIn, isLoading: isAuthenticating } = useAuthenticated();
 
   useEffect(() => {
-    if (isLoggedIn && !Object.keys(response).length) {
+    if (isLoggedIn) {
```
RitikJaiswal75, line 35: since you removed length from the if condition, how are you planning to handle users with 0 tasks?

RitikJaiswal75: Also please update the tests in case they were supposed to

harshith-venkatesh: @RitikJaiswal75 , @kshitij430 has added description addressing it

harshith-venkatesh: Test cases would have helped here

harshith-venkatesh:
Could you add video to showcase the issue being resolved in different scenarios please @kshitij430

src/pages/mine.tsx
Maheima: Can you point us to a resource that can help in writing test for this scenario?

Maheima (dismissed):

harshith-venkatesh:
From contract , response can't be null or undefined or some error?
If we have proper error handling ,we are good with merging

kshitij430:
> From contract , response can't be null or undefined or some error? If we have proper error handling ,we are good with merging

I doubt the API will give a null or undefined but In cases if the response might be null or undefined. It is better to check if there is a required response present and then update the state for tasks.
In a scenario if there is an error the jsx return will handle it.

Maheima (approved):
LGTM!

PR #398, opened 2023-01-12
Refactor/package upgrade

Maheima:
- Major version upgrades: 

| Package name  | Old version | New version |
| ------------- | ------------- | ------------- |
| Node  | 16.x  | 18.x |
| Next  | 12  | 13 |
| Jest | 28.x | 29.x |

- Because of Nextjs upgrade to version 13, Link and Image tag in a few components have been updated.
- Checksum changes when you change the version of your mock service worker package according to what I can conclude from this https://mswjs.io/docs/faq#why-do-i-see-the-detected-outdated-service-worker-error-in-my-console


src/components/Dropdown/Dropdown.tsx
```diff
@@ -17,13 +16,11 @@ const Dropdown = () => {
   return (
     <div className={styles.dropdown}>
       <ul className={styles.dropdownList}>
-        <Link href={USER_PROFILE_URL}>
+        <a href={USER_PROFILE_URL} className={styles.dropdownLink}>
```
RitikJaiswal75, line 19: cant this be done by link tag?

RitikJaiswal75: Next 13 gives an error on using a tag inside the link tag but the link tag by itself is not an issue.

Example to use classNames on link tag: https://github.com/vercel/next.js/blob/9a82bba86ed2fba02c651983c78eac744a4d75d3/examples/active-class-name/components/ActiveLink.tsx#L52

RitikJaiswal75: In case you are doing this so you could place `li` inside `a` you could use this:
https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

src/components/idleUsers/card/index.tsx
```diff
@@ -17,20 +16,18 @@ const Card: FC<Props> = ({ user }) => {
   const profileUrl = `${MEMBERS_URL}/${user.username}`
 
   return (
-    <Link href={profileUrl}>
-      <a data-testid='profile-card' className={styles.card}>
+      <a data-testid='profile-card' className={styles.card} href={profileUrl} >
```
RitikJaiswal75, line 19: Why are we changing every a tag to link tag?

RitikJaiswal75, line 19: Why are we changing every a tag to link tag?

public/mockServiceWorker.js
```diff
@@ -2,13 +2,13 @@
...
 /**
- * Mock Service Worker (0.44.2).
+ * Mock Service Worker (0.49.2).
  * @see https://github.com/mswjs/msw
  * - Please do NOT modify this file.
  * - Please do NOT serve this file on production.
  */
 
-const INTEGRITY_CHECKSUM = 'b3066ef78c2f9090b4ce87e874965995'
+const INTEGRITY_CHECKSUM = '3d6b9f06410d179a7f7404d4bf4c3c70'
```
RitikJaiswal75, line 11: Is this required in this PR?

src/components/idleUsers/card/index.tsx
```diff
@@ -17,20 +16,18 @@ const Card: FC<Props> = ({ user }) => {
...
-    <Link href={profileUrl}>
-      <a data-testid='profile-card' className={styles.card}>
+      <a data-testid='profile-card' className={styles.card} href={profileUrl} > 
         <Image
           src={userImg || DUMMY_PROFILE}
           alt={user.full_name}
           width={150}
           height={150}
           data-testid='user-image'
+          priority={true}
```
RitikJaiswal75, line 26: won't setting a priority make our page load slow(er)?

src/components/navBar/index.tsx
```diff
@@ -35,14 +35,14 @@ const NavBar = () => {
       <div>
         {!isLoggedIn ? (
           <Link href={LOGIN_URL}>
-            <a className={styles.signInLink}>
+            <p className={styles.signInLink}>
               Sign In With GitHub
               <img
```
RitikJaiswal75, line 40: can we use the NextJs Image tag here?

tsconfig.json
```diff
@@ -20,9 +17,7 @@
     "isolatedModules": true,
     "baseUrl": ".",
     "paths": {
-      "@/*": [
-        "src/*"
-      ]
+      "@/*": ["src/*"]
```
RitikJaiswal75, line 20: they are just formatting changes not a part of the package upgrade to refactor.

Maheima: But that is the correct formatting right? Then, why not just add it?

RitikJaiswal75: lets say that is the formatting in your system in my system prettier runs in some different way then?

heyrandhir:
I will approve once Ritik comments are resolved. @Maheima

src/components/idleUsers/card/index.tsx
heyrandhir: yeah also would like to know why are we setting this to true, as far as I read on the docs `You should use the priority property on any image detected as the [Largest Contentful Paint (LCP)](https://nextjs.org/learn/seo/web-performance/lcp) element`. So are these images largest element on the page. If yes then we can go with true else we can default to false.

public/mockServiceWorker.js
Maheima: Yes. Checksum changes when you change the version of your mock service worker package

tsconfig.json
Maheima: This is the formatting specified for our project

RitikJaiswal75:
Can we have a small description here @Maheima

heyrandhir (approved):

samyakshah3008 (approved):
Great work Mahima! 
Looks good to merge ✅

shmbajaj (approved):
Looks good to merge ✅

Issue #395, opened 2023-01-10
Fix CSS of cards in idle-user tab


Maheima:
![Screenshot_20230110-222154](https://user-images.githubusercontent.com/42349041/211613309-e4db4dd0-24ff-4059-a18a-bc2951ecb643.png)

shmbajaj:
Hey @Maheima assign this to me.

samyakshah3008:
I would love to work on this task @Maheima, can you please assign me this task?

PR #401, opened 2023-01-26
Fix/css idle user card

Maheima:
Fixes [395](https://github.com/Real-Dev-Squad/website-status/issues/395)
Handles long names better. Earlier, the height was fluctuating when a long name was there.

## Before:
image

## After:
image


src/components/idleUsers/card/card.module.scss
```diff
@@ -25,9 +25,13 @@ $theme-pink: #e30464;
   }
 
   .name {
-    margin-top: 3%;
-    margin-bottom: 0;
+    display: flex;
     font-size: 1rem;
     font-weight: 700;
+    width: 160px;
```
harshith-venkatesh, line 31: Do we need width and height?

Maheima: Yes, otherwise overflow won't be possible

Maheima: Yes, otherwise the height and width changes depending upon the length of the name

samyakshah3008, line 31: Can we give width in `rem` instead of `px`? As rem is preferrable when value is greater than 20px.

samyakshah3008 (dismissed):

src/components/idleUsers/card/card.module.scss
RitikJaiswal75: werent we going to use `text-overflow: ellipsis;`

Maheima: There wasn't a requirement of it after using flex because the p tag was not centered so, using flex fixed both of the things

Maheima: updated it

src/components/idleUsers/card/card.module.scss
```diff
@@ -25,9 +25,13 @@ $theme-pink: #e30464;
...
 
   .name {
-    margin-top: 3%;
-    margin-bottom: 0;
+    display: flex;
     font-size: 1rem;
     font-weight: 700;
+    width: 10rem;
+    text-align: center;
+    height: 1.5rem;
```
yesyash, line 33: This is a good approach but can cause problems when the username is too big. 
I generally use `min-height` property to set a default height to the element.

src/components/idleUsers/card/card.module.scss
```diff
@@ -25,9 +25,13 @@ $theme-pink: #e30464;
   }
 
   .name {
-    margin-top: 3%;
-    margin-bottom: 0;
+    display: flex;
     font-size: 1rem;
     font-weight: 700;
+    width: 10rem;
```
yesyash, line 31: Instead of setting the width in rem units can we not let the width be the default width (100% of the container) and use padding to reduce the horizontal area? 

This way if a user changes their default scaling from 16px to something bigger, it would not overflow from the card.

yesyash (approved):

src/components/idleUsers/card/card.module.scss
Maheima: But that would get the height of card to increase if we get the user's last name in new line when the first name is long

src/components/idleUsers/card/card.module.scss
Maheima: I tried doing this but it changed the width of the card then

yesyash: Yes, it does, if set 100% the responsibility of handling width should go to the parent. 
Example: [https://zx588j.csb.app/](url)

samyakshah3008 (approved):
Looks good for approval. 
Cool work, Mahima

ivinayakg (commented):
What if we set a fixed width and height of then, in js check wether the the length of name is bigger than "n" (n being the max length of the name which will not break the card design) and if yes then slice the name? 
Wdyt

src/components/idleUsers/card/card.module.scss
Maheima: Is it okay now?

Maheima:
> What if we set a fixed width and height of then, in js check wether the the length of name is bigger than "n" (n being the max length of the name which will not break the card design) and if yes then slice the name? Wdyt

I did that initially but Ankush suggested that we hide overflow instead

src/components/idleUsers/card/card.module.scss
```diff
@@ -13,21 +13,25 @@ $theme-pink: #e30464;
...
 
   .name {
-    margin-top: 3%;
-    margin-bottom: 0;
     font-size: 1rem;
     font-weight: 700;
+    width: 100%;
+    text-align: center;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
heyrandhir, line 34: Nicely handled @Maheima 👍🏻

heyrandhir (approved):

sahsisunny (approved):

src/components/idleUsers/card/card.module.scss
Maheima: Thank you :)

PR #405, opened 2023-01-29
Resolve flash errors

Maheima:
- Earlier the filter API was getting hit even if the user wasn't logged in, which was causing the errors, changed it to be hit only when the user is logged in. 
- Added the `isLoggedIn` returned from the `useAuthenticated`  and passed it as state to the App wrapper context. This way we can use it anywhere we need to check if the user is logged in and then, hit the APIs that need authentication.
- Also, refactored the challenges and mine page to use the state we get from context instead of using useAuthenticated hook for it

## Dev tested: 
- [x] Yes
- [ ] No


src/components/tasks/card/index.tsx
```diff
@@ -39,33 +39,36 @@ const Card: FC<Props> = ({
...
+              }
+            });
+          const { data: result } = await requestPromise
+          setTaskTagLevel(result.data)
+        } catch (err: any) {
+          toast(ERROR, err.message);
+        }
+      })()
+    }
+  }, [])
```
harshith-venkatesh, line 71: Have we checked for dependencies here?
And would it be better to take the function outside the useEffect here?

src/pages/mine.tsx
```diff
@@ -29,8 +29,9 @@ const Mine: FC = () => {
     isLoading,
     callAPI
   } = useFetch(TASKS_URL, {}, false);
-  const { isLoggedIn, isLoading: isAuthenticating } = useAuthenticated();
-
+  const context = useAppContext();
+  const { state } = context || {};
```
harshith-venkatesh, line 33: Wouldn't it be better to initiate the context with {} instead of declaring it everywhere?
Are there any reason for going with this approach?

src/components/tasks/card/index.tsx
```diff
@@ -39,33 +39,36 @@ const Card: FC<Props> = ({
...
+                itemType: 'TASK',
+                itemId: `${cardDetails.id}`
+              }
+            });
+          const { data: result } = await requestPromise
+          setTaskTagLevel(result.data)
+        } catch (err: any) {
+          toast(ERROR, err.message);
+        }
+      })()
```
heyrandhir, line 69: Instead of an IIFE. Can we extract it out to a separate function and call the function here instead.

src/components/tasks/card/index.tsx
```diff
@@ -39,33 +39,36 @@ const Card: FC<Props> = ({
   const [showEditButton, setShowEditButton] = useState(false);
   const [keyLongPressed] = useKeyLongPressed();
   const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const context = useAppContext();
+  const { actions, state } = context || {};
```
heyrandhir, line 44: The term `state` is a fairly general. We could have been more specific like `idleUserState` or a related term.

src/components/tasks/card/index.tsx
heyrandhir, line 71: Shouldn't `state` be in the dependency array here 👀

src/components/tasks/card/index.tsx
```diff
@@ -39,33 +39,36 @@ const Card: FC<Props> = ({
...
-  const context = useAppContext() ;
-  const { actions } = context || {}
+    if (state.isLoggedIn) {
+      (async () => {
+        try {
+          const { requestPromise } = fetch(
+            {
+              url: `${ITEM_BY_FILTER_URL}`,
+              params: {
+                itemType: 'TASK',
```
heyrandhir, line 60: if `ITEM_BY_FILTER_URL` is imported from constant. can we not add even `TASK` into constants.

heyrandhir (changes requested):

src/components/tasks/card/index.tsx
Maheima: This state here refers to all the states in App context

yesyash: What @heyrandhir might be trying to tell here (and something I agree with) is one context should store and mutate only a single type of data. 
Example: 
- userSettings
- cartItems

And hence it should be named specifically to its use case and not a generic term.

src/components/tasks/card/index.tsx
```diff
@@ -39,33 +39,36 @@ const Card: FC<Props> = ({
...
+                itemId: `${cardDetails.id}`
+              }
+            });
```
yesyash, line 63: In Strict mode fetching data without a cleanup function to ignore the request might cause bugs in development mode. 
Attaching a link to the react beta docs where they give a good example of how to fetch data. 

[https://beta.reactjs.org/learn/synchronizing-with-effects#fetching-data](url)

src/components/tasks/card/index.tsx
Maheima: Done

src/pages/mine.tsx
Maheima: Made the change

src/components/tasks/card/index.tsx
Maheima: Because it is a global state and we are checking for login status I don't think there is a need for a separate one, if it is required then, it would rather be a better idea to call the Appwrapper context, editContext. I can create a refactor ticket for it

src/components/tasks/card/index.tsx
Maheima: done

src/components/tasks/card/index.tsx
Maheima: But what if in future we want to add tags to something other than tasks. It wouldn't be generalised then

samyakshah3008 (dismissed):
Looks good to merge ✅

src/components/tasks/card/index.tsx
Maheima: I haven't added this. I have created a refactor ticket for this https://github.com/Real-Dev-Squad/website-status/issues/406 
This component needs a revamp.

src/components/tasks/card/index.tsx
```diff
@@ -39,33 +39,37 @@ const Card: FC<Props> = ({
...
-         toast(ERROR, err.message);
-       }
-    })()
-  },[])
   
-  const context = useAppContext() ;
-  const { actions } = context || {}
+  useEffect(() => {
+    if (state?.isLoggedIn) {
+      const getTaskTags = async () => {
```
harshith-venkatesh, line 54: I feel like we could define the function irrespective of useEffect

Maheima: Added

heyrandhir (approved):

RitikJaiswal75 (approved):
LGTM

samyakshah3008 (approved):
Looks good to merge, great work Mahima!

harshith-venkatesh (approved):

src/context/index.tsx
```diff
@@ -19,10 +20,13 @@ const AppWrapperContext: FC<Props> = ({ children }) => {
     const onEditRoute = () => {
         router.replace('?edit=true'); //TODO handle query better https://github.com/Real-Dev-Squad/website-status/issues/299 
     }
+    const { isLoggedIn, isLoading } = useAuthenticated();
```
ivinayakg, line 23: also take the userData from the hook so that you can also pass that down.

src/pages/challenges.tsx
```diff
@@ -27,7 +27,8 @@ const Challenges: FC = () => {
     error,
     isLoading,
     callAPI } = useFetch(CHALLENGES_URL, {}, false);
-  const { isLoggedIn, isLoading: isAuthenticating } = useAuthenticated();
+  const { state } = useAppContext();
+  const { isLoading: isAuthenticating, isLoggedIn } = state;
 
   useEffect(() => {
```
ivinayakg, line 33: we won't need this effect if we take the userData from the context itself right?

ivinayakg (approved):

PR #416, opened 2023-02-07
Custom server added to fix MSW setup

sahsisunny:
# Custom server added to fix MSW setup

## What needs to be done

- First of all, create the new certificate for the server and add it to the new folder `certificates` in the root of the project. Also add path to the certificate in the `server.js` file.
- Run the server with `yarn run dev:server` which is already added to the `package.json` file.
- Then go to `dev.realdevsquad.com` and you should see the `MOCKING ENABLED` message in the console.


sahsisunny:
New PR - #435

PR #451, opened 2023-03-13
remove hardcoded value of URLs

### LINKED ISSUE #450: Remove hardcoded value of APIs
Currently, we are hardcoding the value of API in some places. Change it to get from `.env.development` file instead

Maheima:
Resolves #450 

- Added the value of URLs to constants
- Added the base URL values from `.env.development` file


kotesh-arya (approved):
Looking good enough for a merge 👍🏻

ruthv1k (approved):

PR #471, opened 2023-04-03
prettier and lint configure

bhtibrewal:


.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  trailingComma: "es5",
+  tabWidth: 4,
+  semi: true,
+  singleQuote: false,
```
prakashchoudhary07, line 5: I personally prefer tab width as 2 and use single quotes

I think we should change this, lot of space is wasted when tab width is 4, and every time we have to type double quotes we have to hit 2 keys, shift + quotes key, where as in single quotes we just hit one key

.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  trailingComma: "es5",
```
prakashchoudhary07, line 2: What does es5 signify here?
When all will it add tralingComma

.eslintrc.json
```diff
@@ -1,11 +1,26 @@
...
+    "env": {
+        "browser": true,
+        "es2021": true,
+        "node": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:react/recommended",
+        "plugin:@typescript-eslint/recommended"
+    ],
```
prakashchoudhary07, line 11: Will this work well with prettire?

.prettierrc.js
bhtibrewal: please read this section for reference on `trailingComma: "es5",` https://prettier.io/docs/en/options.html#trailing-commas

.eslintrc.json
bhtibrewal: I didn't find any issues with this and prettier.

.prettierrc.js
vvaibhavdesai: @bhtibrewal the more common accepted way is single quotes.

also tabWitdth seems alright to me with 4 spaces

bhtibrewal: Thank you @prakashchoudhary07 and @vvaibhavdesai for pointing out these, as you both and mahima said I have made the change singlequote: true and we are going ahead with tabwidth 4

vvaibhavdesai (approved):

prakashchoudhary07 (approved):

PR #473, opened 2023-04-03
Show cloudinary profile image for users

### LINKED ISSUE #470: Status site should use cloudinary images
In the current version of our site, we display images from our "website-static" repository. However, this is not an ideal solution as it requires us to manually upload images to the repository. Instead, we have images stored in Cloudinary, which are automatically uploaded to the platform. It would be better to use these images instead of the ones from our repository.

sahsisunny:
# Show cloudinary profile image for users

### What is the change?

- `src/app/services/usersApi.ts` - Added transformResponse to return the response data
- `src/components/availability-panel/drag-drop-context/DraggableComponent.tsx` - Get the response data from the userApi service and use it to display the profile image.
- `src/components/tasks/card/index.tsx` - Get the response data from the userApi service and use it to display the profile image.

### Is Development Tested?

- [x] Yes
- [ ] No

### After Change Screenshots

image

### DIFF
--- a/src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
+++ b/src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
@@ -5,12 +5,12 @@ import {
 } from 'react-beautiful-dnd';
 import { FC, useContext } from 'react';
 import { draggableProps } from '@/interfaces/availabilityPanel.type';
+import { DUMMY_PROFILE as placeholderImageURL } from '@/components/constants/display-sections';
+import { useGetUsersByUsernameQuery } from '@/app/services/usersApi';
 import classNames from '@/components/availability-panel/drag-drop-context/styles.module.scss';
 import { disableDrag } from '.';
 import Image from 'next/image';
 
-const imageGenerator = (name: string) =>
-    `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${name}/img.png`;
 
 const getItemStyle = (
     isDragging: boolean,
@@ -38,6 +38,12 @@ const DraggableComponent: FC<draggableProps> = ({
     title = '',
 }) => {
     const draggableIds = useContext(disableDrag);
+        const { data: userResponse } = useGetUsersByUsernameQuery({
+        searchString: draggableId,
+        size: 1,
+    });
+    const draggableUserImageURL: string =
+        userResponse?.users[0]?.picture?.url || placeholderImageURL;
     return (
         <Draggable
             key={draggableId}
@@ -62,14 +68,10 @@ const DraggableComponent: FC<draggableProps> = ({
                     ) : (
                         <div style={{ display: 'flex', alignItems: 'center' }}>
                             <Image
-                                src={imageGenerator(draggableId)}
+                                src={draggableUserImageURL}
                                 alt={draggableId}
                                 width={52}
                                 height={52}
-                                onError={(e) => {
-                                    (e.target as HTMLImageElement).src =
-                                        'dummyProfile.png';
-                                }}
                             />
                             <span>{draggableId}</span>
                         </div>
--- a/src/components/tasks/card/index.tsx
+++ b/src/components/tasks/card/index.tsx
@@ -14,6 +14,7 @@ import TaskLevelEdit from './TaskTagEdit';
 import { updateTaskDetails } from '@/interfaces/taskItem.type';
 import fetch from '@/helperFunctions/fetch';
 import { TASKS_URL } from '@/components/constants/url';
+import { DUMMY_PROFILE as placeholderImageURL} from '@/components/constants/display-sections';
 import styles from '@/components/issues/Card.module.scss';
 import moment from 'moment';
 import { Loader } from './Loader';
@@ -23,6 +24,7 @@ import {
     useDeleteTaskTagLevelMutation,
     useGetTaskTagsQuery,
 } from '@/app/services/taskTagApi';
+import { useGetUsersByUsernameQuery } from '@/app/services/usersApi';
 
 type Props = {
     content: task;
@@ -44,9 +46,12 @@ const Card: FC<Props> = ({
         TASK_STATUS.AVAILABLE,
     ];
     const cardDetails = content;
-    const [assigneeProfilePic, setAssigneeProfilePic] = useState(
-        `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${cardDetails.assignee}/img.png`
-    );
+        const { data: userResponse } = useGetUsersByUsernameQuery({
+        searchString: cardDetails.assignee,
+        size: 1,
+    });
+    const assigneeProfileImageURL: string =
+        userResponse?.users[0]?.picture?.url || placeholderImageURL;
     const { SUCCESS, ERROR } = ToastTypes;
     const isUserAuthorized = useContext(isUserAuthorizedContext);
 
@@ -76,9 +81,6 @@ const Card: FC<Props> = ({
             setShowEditButton(true);
         }
     }, [keyLongPressed]);
-
-    const contributorImageOnError = () =>
-        setAssigneeProfilePic('/dummyProfile.png');
 
     const localStartedOn = new Date(parseInt(cardDetails.startedOn, 10) * 1000);
     const fromNowStartedOn = moment(localStartedOn).fromNow();
@@ -430,9 +432,8 @@ const Card: FC<Props> = ({
                         </span>
                         <span className={classNames.contributorImage}>
                             <Image
-                                src={assigneeProfilePic}
-                                alt={`profile picture of ${cardDetails.assignee}`}
-                                onError={contributorImageOnError}
+                                src={assigneeProfileImageURL}
+                                alt={cardDetails.assignee || 'dummy profile'}
                                 width={30}
                                 height={30}
                             />
@@ -586,9 +587,8 @@ const Card: FC<Props> = ({
                             </span>
                             <span className={classNames.contributorImage}>
                                 <Image
-                                    src={assigneeProfilePic}
-                                    alt="Assignee profile picture"
-                                    onError={contributorImageOnError}
+                                    src={assigneeProfileImageURL}
+                                    alt={cardDetails.assignee || 'dummy profile'}
                                     width={45}
                                     height={45}
                                 />


src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -33,6 +32,13 @@ const DraggableComponent: FC<draggableProps> = ({
   title = '',
 }) => {
   const draggableIds = useContext(disableDrag);
+  const [draggableImage, setDraggableImage] = useState<string>('');
+
+  useEffect(() => {
+      getAssigneeProfilePic(draggableId).then((res) => {
+        setDraggableImage(res);
+      });
+  }, [])
```
bhtibrewal, line 41: This API will also be called when the component unmounts, this will hit the performance

src/components/tasks/card/index.tsx
```diff
@@ -53,6 +52,11 @@ const Card: FC<Props> = ({
     if (state?.isLoggedIn) {
       getTaskTags();
     }
+    if (cardDetails.assignee) {
+      getAssigneeProfilePic(cardDetails.assignee).then((res) => {
```
hjaintech, line 56: [nitpick] Wondering if there is a specific reason for NOT using `await` here?

src/components/tasks/card/index.tsx
```diff
@@ -30,9 +31,7 @@ const Card: FC<Props> = ({
   const statusRedList = [BLOCKED];
   const statusNotOverDueList = [COMPLETED, VERIFIED, AVAILABLE];
   const cardDetails = content;
-  const [assigneeProfilePic, setAssigneeProfilePic] = useState(
-    `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${cardDetails.assignee}/img.png`,
-  );
+  const [assigneeProfilePic, setAssigneeProfilePic] = useState<string>('');
```
hjaintech, line 34: [nitpick] Do we need to specify the type of state value since the default value would anyways imply to string?

src/components/tasks/card/index.tsx
```diff
@@ -53,6 +52,11 @@ const Card: FC<Props> = ({
     if (state?.isLoggedIn) {
       getTaskTags();
     }
+    if (cardDetails.assignee) {
```
hjaintech, line 55: Would you want to do this only when user is logged in? If yes, then should this check `state.isLoggedIn` as well?

src/components/tasks/card/index.tsx
```diff
@@ -325,8 +329,8 @@ const Card: FC<Props> = ({
             className={classNames.contributorImage}
           >
             <Image
-              src={assigneeProfilePic}
-              alt="Assignee profile picture"
+              src={assigneeProfilePic || '/dummyProfile.png'}
```
hjaintech, line 332: Would it be better to keep the default value of `assigneeProfilePic` state as `/dummyProfile.png` instead of short-circuting it here? Same for `alt` text as well.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -33,6 +32,13 @@ const DraggableComponent: FC<draggableProps> = ({
   title = '',
 }) => {
   const draggableIds = useContext(disableDrag);
+  const [draggableImage, setDraggableImage] = useState<string>('');
```
hjaintech, line 35:
```suggestion
  const [draggableImageUrl, setDraggableImageUrl] = useState<string>('');
```

src/helperFunctions/getAssigneeProfilePic.ts
```diff
@@ -0,0 +1,14 @@
+import { USERS } from '@/components/constants/url';
+import fetch from '@/helperFunctions/fetch';
+
+const getAssigneeProfilePic = async (username: string) => {
+     try {
+          const url = `${USERS}/${username}`;
+          const { requestPromise } = fetch({ url });
+          const { data: result } = await requestPromise
+          return result.user?.picture?.url
+     } catch (err: any) {
```
shmbajaj, line 10: Why error is not handled here?

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -57,8 +63,8 @@ const DraggableComponent: FC<draggableProps> = ({
             : (
               <div style={{ display: 'flex', alignItems: 'center' }}>
                 <Image
-                  src={imageGenerator(draggableId)}
-                  alt={draggableId}
+                  src={draggableImage || '/dummyProfile.png'}
```
shmbajaj, line 66: Store `dummy image url` in a constant.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -33,6 +32,13 @@ const DraggableComponent: FC<draggableProps> = ({
   title = '',
 }) => {
   const draggableIds = useContext(disableDrag);
+  const [draggableImage, setDraggableImage] = useState<string>('');
+
+  useEffect(() => {
+      getAssigneeProfilePic(draggableId).then((res) => {
+        setDraggableImage(res);
```
shmbajaj, line 39: Here please check for error and response, also is response a url?

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
shmbajaj: @bhtibrewal are you sure API will be called again when component `umounts`, not on mounting-again?

src/components/tasks/card/index.tsx
shmbajaj: +1

shmbajaj (changes requested):

src/components/tasks/card/index.tsx
sahsisunny: Yes you are right there is  no need to specify that. I will remove it

src/components/tasks/card/index.tsx
sahsisunny: I'll put it in a new useEffect.

src/components/tasks/card/index.tsx
sahsisunny: ok i will remove it.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: Done

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: yes response is URL

src/helperFunctions/getAssigneeProfilePic.ts
sahsisunny: done

src/components/tasks/card/index.tsx
sahsisunny: We can't wait for the image to load, then load the task. First the task gets loaded then the images keep loading in the background

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: i am not able to see @bhtibrewal  in network section can you please send me the videos/images ?

src/helperFunctions/getAssigneeProfilePic.ts
```diff
@@ -0,0 +1,16 @@
...
+import fetch from '@/helperFunctions/fetch';
+import { IMAGE_NOT_FOUND_ERROR_MESSAGE } from '@/components/constants/messages'
+
+const getAssigneeProfilePic = async (username: string) => {
+     try {
+          const url = `${USERS}/${username}`;
+          const { requestPromise } = fetch({ url });
+          const { data: result } = await requestPromise
+          return result.user?.picture?.url
+     } catch (err: any) {
```
heyrandhir, line 11: can we have type for err as well instead of `any`.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -33,6 +33,14 @@ const DraggableComponent: FC<draggableProps> = ({
   title = '',
 }) => {
   const draggableIds = useContext(disableDrag);
+  const [draggableImageUrl, setDraggableImageUrl] = useState(DUMMY_PROFILE);
+  const draggableImageOnError = useCallback(() => setDraggableImageUrl(DUMMY_PROFILE), []);
```
vvaibhavdesai, line 37: do we really need to use `useCallback` here for setting a state? as I don't see a reason to add setState in useCallback. Also, the dependency array is empty

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -33,6 +33,14 @@ const DraggableComponent: FC<draggableProps> = ({
...
+
+  useEffect(() => {
+    getAssigneeProfilePic(draggableId).then((res) => {
```
vvaibhavdesai, line 40: why is `.catch` missing?

also in case of `res` comes empty how are we handling that?

src/components/tasks/card/index.tsx
vvaibhavdesai: Your image will load async and your code will be loaded by the time the image loads which means your component will be visible to the user. images won't load in that but the `res` of the component will load. 

also, the video is recorded in fast 3g throttled mode.

I will request you to make changes and test once and let me know if it creates any issue


https://user-images.githubusercontent.com/64726586/230130458-1530c793-3e33-4134-a845-62c2adbf1e06.mp4

src/helperFunctions/getAssigneeProfilePic.ts
```diff
@@ -0,0 +1,16 @@
+import { USERS } from '@/components/constants/url';
+import fetch from '@/helperFunctions/fetch';
+import { IMAGE_NOT_FOUND_ERROR_MESSAGE } from '@/components/constants/messages'
+
+const getAssigneeProfilePic = async (username: string) => {
+     try {
+          const url = `${USERS}/${username}`;
+          const { requestPromise } = fetch({ url });
```
vvaibhavdesai, line 8: why awaiting on `requestPromise `and not on `fetch `directly?

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
shmbajaj, line 37: You don't need to use `useCallback` here, as the cost to save recreation of the onErrorHandler function is not big enough.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
shmbajaj: +1

src/components/tasks/card/index.tsx
shmbajaj: @vvaibhavdesai I guess this is what he want to do load images async, and a show placeholder meanwhile.

Correct me If a wrong @sahsisunny

src/components/tasks/card/index.tsx
```diff
@@ -72,7 +80,7 @@ const Card: FC<Props> = ({
     }
   }
 
-  const contributorImageOnError = () => setAssigneeProfilePic('/dummyProfile.png');
+  const contributorImageOnError = useCallback(() =>  setAssigneeProfilePic(DUMMY_PROFILE), []);
```
shmbajaj, line 83: Check the above `useCallback` comments.

shmbajaj (changes requested):

heyrandhir (changes requested):

src/components/tasks/card/index.tsx
```diff
@@ -55,6 +55,14 @@ const Card: FC<Props> = ({
     }
   }, [state?.isLoggedIn])
 
+  useEffect(() => {
+    if (cardDetails.assignee) {
```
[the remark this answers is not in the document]
sahsisunny: assignee is string type in data-modal

src/helperFunctions/getAssigneeProfilePic.ts
sahsisunny: The **requestPromise** is destructured and awaited instead of the fetch Promise to asynchronously process the response data.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: Using `useCallback`,  we are ensuring that the function is memoized and only gets redefined if the **draggableId** changes. This helps to prevent unnecessary re-renders and optimize the performance of the component. its a small improvement but @shmbajaj  said its not necessary then i will remove it.

src/components/tasks/card/index.tsx
sahsisunny: > @vvaibhavdesai I guess this is what he want to do load images async, and a show placeholder meanwhile.
> 
> Correct me If a wrong @sahsisunny

Yes

src/helperFunctions/getAssigneeProfilePic.ts
```diff
@@ -0,0 +1,16 @@
...
+          const { data: result } = await requestPromise
```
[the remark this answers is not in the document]
sahsisunny: Screenshot (144)

src/components/tasks/card/index.tsx
vvaibhavdesai: @shmbajaj that can be done irrespective. Nothing to do with awaiting here.
Also, this is what he said 
image
and the video I shared is the code written with `async` and `await`

shmbajaj (approved):
LGTM ✅

hjaintech (dismissed):

DeRaowl (dismissed):
LGTM

heyrandhir (approved):

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -38,6 +37,20 @@ const DraggableComponent: FC<draggableProps> = ({
...
+        setdraggableUserImageUrl(DUMMY_PROFILE);
+
+    useEffect(() => {
+        getUserProfileImage(draggableId)
+            .then((res) => {
+                setdraggableUserImageUrl(res);
+            })
+            .catch(() => {
+                setdraggableUserImageUrl(DUMMY_PROFILE);
+            });
```
ankushdharkar, line 52: Does this mean that if we have a 100 tasks assigned to the **same** user, then we would make 100 image request calls to the same URL via `getUserProfileImage`? That seems wasteful to me.

Should we think about leveraging a global store with some caching enabled to reduce these unnecessary calls?

src/components/tasks/card/index.tsx
```diff
@@ -64,6 +64,18 @@ const Card: FC<Props> = ({
         }
     }, [state?.isLoggedIn]);
 
+    useEffect(() => {
```
ankushdharkar, line 67: Using side-effects for this feels like bad design. Can we improve it using a global store like mentioned above?

src/helperFunctions/getUserProfileImage.ts
```diff
@@ -0,0 +1,18 @@
...
+import { IMAGE_NOT_FOUND_ERROR_MESSAGE } from '@/components/constants/messages';
+
+const getUserProfileImage = async (username: string) => {
+    try {
+        const url = `${USERS}/${username}`;
+        const { requestPromise } = await fetch({ url });
+        const { data } = await requestPromise;
+        return data.user?.picture?.url ?? DUMMY_PROFILE;
+    } catch (err) {
+        console.log(IMAGE_NOT_FOUND_ERROR_MESSAGE);
```
ankushdharkar, line 13: Should this be `console.error`?

ankushdharkar (changes requested):

src/helperFunctions/getUserProfileImage.ts
sahsisunny: done

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
hjaintech: Also, worth checking if this asset would automatically get cached on the browser? After the first load, would subsequent images be served from the browser cache directly or make a fresh network call?

src/components/tasks/card/index.tsx
hjaintech: Worth exploring, what if we put the URL directly in the `img` tag?

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
shmbajaj: > Also, worth checking if this asset would automatically get cached on the browser? After the first load, would subsequent images be served from the browser cache directly or make a fresh network call?

This will be done and maintained by ReduxToolkit. So, it's not required to be done manually.

shmbajaj: > Does this mean that if we have a 100 tasks assigned to the **same** user, then we would make 100 image request calls to the same URL via `getUserProfileImage`? That seems wasteful to me.
> 
> Should we think about leveraging a global store with some caching enabled to reduce these unnecessary calls?

@ankushdharkar which approach you are referring to for caching data manually caching using useEffect, or using a tool like Redux Toolkit with RTK Query for automatic caching?

src/components/tasks/card/index.tsx
ankushdharkar: @hjaintech I believe the Tasks API only returns `userIds`. The user data has to be fetched separately via the UsersAPI to get profile picture for the user.
So, from the list of tasks, we need to find the list of unique userIds, and fetch user data for each userId via Users API.

Pratiyushkumar:
LGTM

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -38,6 +37,13 @@ const DraggableComponent: FC<draggableProps> = ({
     title = '',
 }) => {
     const draggableIds = useContext(disableDrag);
+    const { data: userResponse } = useGetUsersByUsernameQuery({
+        searchString: draggableId,
+        size: 1,
+    });
+    const draggableUserImage: string =
```
shmbajaj, line 44: Suggestion: Renamed it to `draggableUserImageURL`.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -1,17 +1,16 @@
+import { FC, useContext } from 'react';
 import {
     Draggable,
     DraggingStyle,
     NotDraggingStyle,
 } from 'react-beautiful-dnd';
-import { FC, useContext } from 'react';
 import { draggableProps } from '@/interfaces/availabilityPanel.type';
 import classNames from '@/components/availability-panel/drag-drop-context/styles.module.scss';
+import { DUMMY_PROFILE } from '@/components/constants/display-sections';
```
shmbajaj, line 9: Use alias name `placeholderImageURL` for `DUMMY_PROFILE`

Snippet:

```js
import { DUMMY_PROFILE as placeHolderImageURL } from '@/components/constants/display-sections';
```

src/components/tasks/card/index.tsx
```diff
@@ -12,6 +13,7 @@ import {
     COMPLETED,
     VERIFIED,
 } from '@/components/constants/beautified-task-status';
+import { DUMMY_PROFILE } from '@/components/constants/display-sections';
```
shmbajaj, line 16: Suggestion:

Use alias name `placeholderImageURL` for `DUMMY_PROFILE`

Snippet:

```js
import { DUMMY_PROFILE as placeHolderImageURL } from '@/components/constants/display-sections';
```

shmbajaj (commented):
Why `Provider` and `Store` is added to test files?

sahsisunny:
> Why `Provider` and `Store` is added/changes-made to test files?

Because we using redux store for fetch image
Screenshot (152)

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: Changes done

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: Changes done

src/components/tasks/card/index.tsx
sahsisunny: Changes done

src/components/tasks/card/index.tsx
```diff
@@ -39,9 +41,13 @@ const Card: FC<Props> = ({
...
     const statusNotOverDueList = [COMPLETED, VERIFIED, AVAILABLE];
     const cardDetails = content;
-    const [assigneeProfilePic, setAssigneeProfilePic] = useState(
-        `${process.env.NEXT_PUBLIC_GITHUB_IMAGE_URL}/${cardDetails.assignee}/img.png`
-    );
+    const { data: userResponse } = useGetUsersByUsernameQuery({
+        searchString: cardDetails.assignee,
+        size: 1,
+    });
+    const assigneeProfileImage: string =
```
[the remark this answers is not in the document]
sahsisunny: Changes done

bhtibrewal (approved):
LGTM

src/context/index.tsx
```diff
@@ -18,7 +18,7 @@ const AppWrapperContext: FC<Props> = ({ children }) => {
     const { edit: editQuery } = query;
     const isEditMode = editQuery === 'true';
     const onEditRoute = () => {
-        router.push({ query: 'edit=true' }, undefined, { shallow: true }); 
+        router.push({ query: 'edit=true' }, undefined, { shallow: true });
```
shmbajaj, line 21: Check the below snippet, and reference url for more info.

```ts
router.push({ query: { edit: 'true' } }, undefined, { shallow: true });

```

Ref: https://nextjs.org/docs/api-reference/next/router#with-url-object

bhtibrewal (approved):
Looks good to me

src/components/tasks/card/index.tsx
```diff
@@ -586,9 +587,8 @@ const Card: FC<Props> = ({
                             </span>
                             <span className={classNames.contributorImage}>
                                 <Image
-                                    src={assigneeProfilePic}
-                                    alt="Assignee profile picture"
-                                    onError={contributorImageOnError}
+                                    src={assigneeProfileImageURL}
+                                    alt={cardDetails.assignee || 'dummy profile'}
```
DeRaowl, line 591: Can we store dummy profile in constants?

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -5,12 +5,12 @@ import {
 } from 'react-beautiful-dnd';
 import { FC, useContext } from 'react';
 import { draggableProps } from '@/interfaces/availabilityPanel.type';
+import { DUMMY_PROFILE as placeholderImageURL } from '@/components/constants/display-sections';
+import { useGetUsersByUsernameQuery } from '@/app/services/usersApi';
```
DeRaowl, line 9: Let us maintain the import order.
Component import should come after external library import ie ```classnames```

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
```diff
@@ -38,6 +38,12 @@ const DraggableComponent: FC<draggableProps> = ({
     title = '',
 }) => {
     const draggableIds = useContext(disableDrag);
+        const { data: userResponse } = useGetUsersByUsernameQuery({
+        searchString: draggableId,
+        size: 1,
```
DeRaowl, line 43: What do 1 refer?

src/components/tasks/card/index.tsx
```diff
@@ -14,6 +14,7 @@ import TaskLevelEdit from './TaskTagEdit';
 import { updateTaskDetails } from '@/interfaces/taskItem.type';
```
DeRaowl, line 14: We are just stuffing code into this File. Can we think of breaking down the component into smaller once?

DeRaowl (changes requested):

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: The number **1** in the code refers to the size parameter which return only one first response and  **draggableId** is the username parameter passed to the `useGetUsersByUsernameQuery `hook. This hook appears to be part of an API service created using the @reduxjs/toolkit library.

src/components/tasks/card/index.tsx
sahsisunny: Yes, it's a good idea to break down the component into smaller pieces if it becomes too complex. Can i create the new ticket for it ?

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
sahsisunny: done

src/components/tasks/card/index.tsx
DeRaowl: Yes, we can track it in new ticket.

src/components/availability-panel/drag-drop-context/DraggableComponent.tsx
DeRaowl: But if a new developer comes and checks the code they won't get this right? Better to store it in variable

sahsisunny: doen i create the new constants **MAX_SEARCH_RESULTS** and set value 1 for **size**.

DeRaowl (approved):
LGTM

PR #500, opened 2023-04-15
write tasks API using RTK query

bhtibrewal:
**Issue: #495**

### DIFF
--- a/src/app/services/tasksApi.ts
+++ b/src/app/services/tasksApi.ts
@@ -1,0 +1,30 @@
+import task from '@/interfaces/task.type';
+import { api } from './api';
+
+type TasksQueryResponse = {
+    message: string;
+    tasks: task[];
+};
+
+export const tasksApi = api.injectEndpoints({
+    endpoints: (builder) => ({
+        getAllTasks: builder.query<TasksQueryResponse, unknown>({
+            query: () => '/tasks',
+            providesTags: ['Tasks'],
+            transformResponse: (response: TasksQueryResponse) => {
+                return {
+                    ...response,
+                    tasks: response.tasks.sort(
+                        (a: task, b: task) => +a.endsOn - +b.endsOn
+                    ),
+                };
+            },
+        }),
+        getMineTasks: builder.query<TasksQueryResponse, unknown>({
+            query: () => '/tasks/mine',
+            providesTags: ['Tasks'],
+        }),
+    }),
+});
+
+export const { useGetAllTasksQuery, useGetMineTasksQuery } = tasksApi;
--- a/src/app/services/tasksApi.ts
+++ b/src/app/services/tasksApi.ts
@@ -8,7 +8,7 @@ type TasksQueryResponse = {
 
 export const tasksApi = api.injectEndpoints({
     endpoints: (builder) => ({
-        getAllTasks: builder.query<TasksQueryResponse, unknown>({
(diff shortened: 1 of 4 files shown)


src/app/services/tasksApi.ts
```diff
@@ -0,0 +1,30 @@
...
+                    ),
+                };
+            },
+        }),
+        getMineTasks: builder.query<TasksQueryResponse, void>({
+            query: () => '/tasks/mine',
+            providesTags: ['Tasks'],
+        }),
+    }),
+});
```
pallabez, line 28: Nice

src/pages/index.tsx
```diff
@@ -73,8 +47,7 @@ async function updateCardContent(id: string, cardDetails: task) {
 
 const Index: FC = () => {
     const { state: appState } = useAppContext();
-    const [filteredTask, setFilteredTask] = useState<any>([]);
-    const { response, error, isLoading } = useFetch(TASKS_URL);
+    const { data: response, isError, isLoading } = useGetAllTasksQuery();
```
pallabez, line 50:
```suggestion
    const { data: tasks, isError, isLoading } = useGetAllTasksQuery();
```

src/pages/index.tsx
```diff
@@ -83,30 +56,17 @@ const Index: FC = () => {
...
+    const tasksGroupedByStatus = updateTasksStatus(
+        response?.tasks || []
+    ).reduce((acc: Record<string, task[]>, curr: task) => {
+        return acc[curr.status as keyof task]
+            ? {
+                  ...acc,
+                  [curr.status]: [...acc[curr.status as keyof task], curr],
+              }
+            : { ...acc, [curr.status]: [curr] };
+    }, {});
```
pallabez, line 70: Thanks for the refactor

bhtibrewal: Will address that in further iterations, Bharti has also wrote some util function for this, I will check it with that in the next PR

src/pages/index.tsx
bhtibrewal: type TasksQueryResponse = {
    message: string;
    tasks: task[];
};
it also has message, that's why I am using response

pallabez: We don't message right? We can remove that in transformResponse.

src/pages/index.tsx
pallabez: Got it.

src/pages/index.tsx
bhtibrewal: done

src/app/services/tasksApi.ts
bhtibrewal: thanks

src/pages/index.tsx
```diff
@@ -136,18 +96,18 @@ const Index: FC = () => {
             <Head title="Tasks" />
             <TasksProvider>
                 <div className={classNames.container}>
-                    {!!error && <p>{TASKS_FETCH_ERROR_MESSAGE}</p>}
+                    {isError && <p>{TASKS_FETCH_ERROR_MESSAGE}</p>}
```
vvaibhavdesai, line 99: the optional chaining looks too much nested please fix this in a different PR

bhtibrewal: okay

pallabez (approved):
LGTM

vvaibhavdesai (approved):

PR #506, opened 2023-04-20
Tags api rtk query

bhtibrewal:
Issue: #495 

Wrote tasksApi and used it in TaskTagList.tsx
Deleted file tasks.context.tsx that was fetching tags and levels


src/pages/index.tsx
```diff
@@ -94,25 +93,24 @@ const Index: FC = () => {
...
-                            ) : (
-                                !isError && <p>{NO_TASKS_FOUND_MESSAGE}</p>
-                            )}
-                        </>
-                    )}
-                </div>
-            </TasksProvider>
+
+            <div className={classNames.container}>
+                {isError && <p>{TASKS_FETCH_ERROR_MESSAGE}</p>}
```
vinayak-trivedi, line 98: is it working? it returns `error`, not `isError`, check [here](https://redux-toolkit.js.org/rtk-query/overview#configure-the-store)

bhtibrewal: yes @vinayak-trivedi it is working 
image

vinayak-trivedi (approved):
LGTM

src/pages/index.tsx
vvaibhavdesai, line 98: make this something like 
```

let data ;
if (isError){
	data = TASK_FETCH_ERROR_MESSAGE
}
 else if(isLoading){
	data = <p>Loading,/p>
}
else{
	" whatever conditions you want '
}

return(<>{data}</>)

```

bhtibrewal: done

src/pages/index.tsx
```diff
@@ -91,28 +90,24 @@ const Index: FC = () => {
...
+
+    const renderContent = () => {
+        if (isError) return <p>{TASKS_FETCH_ERROR_MESSAGE}</p>;
+
+        if (isLoading) return <p>Loading...</p>;
+
+        return (
+            <div className={classNames.tasksContainer}>
+                {renderTabSection()}
+                {renderTaskList()}
```
pallabez, line 102: Any reason to use function call instead of react component?

bhtibrewal: no thanks for pointing out, let me correct it

yesyash (dismissed):

src/pages/index.tsx
bhtibrewal: done

pallabez (approved):
LGTM

src/pages/index.tsx
```diff
@@ -78,41 +77,35 @@ const Index: FC = () => {
...
-    );
-    return (
-        <Layout>
-            <Head title="Tasks" />
-            <TasksProvider>
-                <div className={classNames.container}>
-                    {isError && <p>{TASKS_FETCH_ERROR_MESSAGE}</p>}
-                    {isLoading ? (
-                        <p>Loading...</p>
+    const Content = () => {
```
vvaibhavdesai, line 80: if am not wrong you have declared a react component inside a react component right ?

bhtibrewal: yes but to take it out I have to pass a few props and move the component to another file, the refactoring will go beyond the scope of this PR so I prefer not to do the refactoring in this PR.

vvaibhavdesai: please make that react component a javascript function it's not good to make a react component inside another react component

bhtibrewal: No, Vaibhav just because it's inside a react component doesn't mean we should make it a function, instead the change required here is we should move it out to a separate file.

vvaibhavdesai: react components are recreated on each render why do we need to have a react component inside another react component? this is a bad practice in the first place. also as you say the change required is to take it out in a separate file then why wait for the refactor PR ?

bhtibrewal: done

vvaibhavdesai (approved):

Issue #481, opened 2023-04-09
Edit mode, add option to select task status


prakashchoudhary07:
Currently, to update the status in edit mode, the status is to be typed manually.
It has to be changed to drop-down options, where the user can select from a set of predefined options and update the task status

PR #510, opened 2023-04-23
Add drop down for superuser edit mode

prakashchoudhary07:
Issue link: #481 

Provide dropdown for superuser to edit task status

##### Before:

image

##### After:

image

image

### DIFF
--- a/src/components/tasks/card/card.module.scss
+++ b/src/components/tasks/card/card.module.scss
@@ -162,6 +162,15 @@ $seeMoreTasksShadow: #595959cc;
     cursor: pointer;
 }
 
+.taskStatusAndDateContainer {
+    display: flex;
+    justify-content: space-between;
+}
+
+.taskStatusEditMode {
+    margin-top: 0.8rem;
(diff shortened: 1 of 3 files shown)


RitikJaiswal75 (commented):
arent we missing tests here?

RitikJaiswal75:
Do we need superuser doing all the status changes, I think assigned, done and unassigned would suffice for superuser. other status changes can be done by developers themselves.

prakashchoudhary07:
> arent we missing tests here?

New PR 😆 
https://github.com/Real-Dev-Squad/website-status/issues/639

prakashchoudhary07:
> Do we need superuser doing all the status changes, I think assigned, done and unassigned would suffice for superuser. other status changes can be done by developers themselves.

Superuser should have the option to change all of it, I think. 
Most of the status will be automated in future

We can discuss about this with the super user and, later change if they tell reduce it

RitikJaiswal75 (approved):

Issue #485, opened 2023-04-10
Navbar in the status site is not responsive


Pratiyushkumar:
When the screen size reduces below 806px then the navbar responsiveness gets disturbed.

This is when the screen size is above 806px, the screen looks like this
![image](https://user-images.githubusercontent.com/82165483/230937917-95a2e724-4586-4548-b080-9808eb4877b3.png)

when the screen size comes below 806 px, the screen looks like this
![image](https://user-images.githubusercontent.com/82165483/230938254-0bbb11f9-60d7-4352-8ee8-674867cde4b0.png)

The solution recommended - is the same as responsive on the member's app.
![image](https://user-images.githubusercontent.com/82165483/230938697-bbd0282d-f0aa-4b15-82cf-4683e669ccea.png)

Pratiyushkumar:
The PR https://github.com/Real-Dev-Squad/website-status/pull/516 was meged

PR #516, opened 2023-04-25
Navbar in the status site is not responsive

Pratiyushkumar:
Issue
https://github.com/Real-Dev-Squad/website-status/issues/485
When the screen size reduces below 806px then the navbar responsiveness gets disturbed.
This is when the screen size is above 806px, the screen looks like this
image
image

What is the change?
On the mobile screen hamburger icon should come and all the navigation links should come inside the menu bar

Is Development Tested?
 Yes
 No
Before / After Change Screenshots
For visual or interaction changes. Please provide a video/screenshot.

https://user-images.githubusercontent.com/82165483/234887612-c3a6627d-db37-424d-97a0-7cf7b4ec0300.mp4


src/components/navBar/navBar.module.scss
```diff
@@ -79,4 +78,107 @@ $theme-active: #87d870;
...
+        text-decoration-thickness: $thickness;
+        text-underline-offset: $offset;
+    }
+}
+@media (max-width: 970px) {
+    .navBar {
+        display: flex;
+        flex-direction: row;
+        flex-wrap: wrap;
+        // border: 2px solid red;
```
bhtibrewal, line 106: remove commented code

src/components/navBar/navBar.module.scss
```diff
@@ -79,4 +78,107 @@ $theme-active: #87d870;
...
+    }
+
+    .hamburgerIcon {
+        display: inline-block !important;
+        height: 73px;
+        width: 73px;
+        padding: 18px;
+        cursor: pointer;
+        font-size: 20px;
+        color: white;
```
bhtibrewal, line 135: use responsive units like rem and variable for colors

Pratiyushkumar: ok will fix it

src/components/navBar/navBar.module.scss
Pratiyushkumar: ok will fix it

src/components/navBar/navBar.module.scss
```diff
@@ -79,4 +78,106 @@ $theme-active: #87d870;
...
+        top: 10px;
+        right: 20px;
+    }
+
+    .showLogo {
+        display: none !important;
+    }
+
+    .hamburgerIcon {
+        display: inline-block !important;
```
bhtibrewal, line 128: I still see important being used at multiple places

src/components/navBar/navBar.module.scss
```diff
@@ -79,4 +78,106 @@ $theme-active: #87d870;
...
+    }
+
+    .navBarMenu li {
+        padding: 12px 40px;
+        font-size: 16px;
+        text-align: left;
+        font: 700 16px Roboto, sans-serif;
+    }
+
+    .navBarMenu li a {
```
bhtibrewal, line 160: use classes instead of this ul, li, a

src/components/navBar/navBar.module.scss
Pratiyushkumar: !important here at hamburgerIcon class is necessary, if I remove this then the hamburger Icon is getting removed and Navbar is not looking good, rest I have removed important from everywhere.

src/components/navBar/navBar.module.scss
Pratiyushkumar: Need to use element selectors, because li and a tag are repetitive in the code and If I use className for li and a tag then it will make the HTML file look boring and dull, I think using element selectors can be useful sometimes in these conditions.

bhtibrewal (approved):
Looks good to me

src/components/navBar/navBar.module.scss
```diff
@@ -79,4 +78,106 @@ $theme-active: #87d870;
...
+        top: 10px;
+        right: 20px;
+    }
+
+    .showLogo {
+        display: none;
+    }
+
+    .hamburgerIcon {
+        display: inline-block !important;
```
ivinayakg, line 128: do remove the !important.

src/components/navBar/navBar.module.scss
```diff
@@ -79,4 +78,106 @@ $theme-active: #87d870;
         color: $white;
         font-weight: 600;
     }
+
+    .hamburgerIcon {
+        display: none;
+    }
+    ul li {
```
ivinayakg, line 85: add this behind a class please.

ivinayakg (changes requested):
do look over the comments, please.
Good Work though!

src/components/navBar/navBar.module.scss
Pratiyushkumar: !important here at hamburgerIcon class is necessary, if I remove this then the hamburger Icon is getting removed and Navbar is not looking good, rest I have removed important from everywhere. I know important should be avoided as much as possible but here it was needed

src/components/navBar/navBar.module.scss
Pratiyushkumar: Need to use element selectors, because li and a tag are repetitive in the code and If I use className for li and a tag then it will make the HTML file look boring and dull, I think using element selectors can be useful sometimes in these conditions.

ivinayakg: this is navbar css, so the CSS applied here will be available throughout the page, leading to problems.
please do the changes.

src/components/navBar/navBar.module.scss
ivinayakg: not a great practice, this is used in cases where we haven't written good CSS, it works as an escape.
Please changes this.

src/components/navBar/navBar.module.scss
Pratiyushkumar: Vinayak, we actually need to discuss this, please let me know your availability so that we can connect to make you understand the scenario

src/components/navBar/navBar.module.scss
Pratiyushkumar: Vinayak, we actually need to discuss this, please let me know your availability so that we can connect to make you understand the scenario

Pratiyushkumar: actually, this is the reason behind to use !important "not a great practice, this is used in cases where we haven't written good CSS, and it works as an escape" because when media query is not applied hamburgericon has display as none but when media query is getting applied  on removing !important same display none property is getting applied which is disrupting navbar so I was not getting options rather than to include !important

Pratiyushkumar: and for further clarifications, let's connect over VC

src/components/navBar/navBar.module.scss
Pratiyushkumar: Vinayak if you see the HTML file in that li, a, is repeated and if I give them the class name then that would make code look worse. that's why I chose to use element selectors instead of class selectors.
and for further clarifications, let's connect over VC

src/components/navBar/navBar.module.scss
```diff
@@ -79,4 +78,104 @@ $theme-active: #87d870;
...
+            cursor: pointer;
+            font-size: 20px;
+            color: $white;
+        }
+    }
+    .signInLink {
+        position: absolute;
+        top: 10px;
+        right: 20px;
+    }
```
Pratiyushkumar, line 129: this .signInLink class is for the sign-in with GitHub button when the user is not logged in and

ivinayakg (approved):

src/components/navBar/navBar.module.scss
ivinayakg: explain this further please.

bhtibrewal (approved):
Okay to merge

src/components/navBar/navBar.module.scss
Pratiyushkumar: This is basically for sign in button and this is given position absolute because it was getting removed from its position when I was trying to fix the navbar so that's why I did this.

PR #517, opened 2023-04-26
optimize api with rtk and added test

arpit01923:


shubhamsigdar1:
> Description -
> 
> 1. Instead of calling API in **useContext**, will add API to **redux toolkit** which do not make unnecessary calls and make it optimized.
> 2. Followed **TDD** approach

Nit:
Can you please connect the issue you are working on?

shubhamsigdar1 (approved):
LGTM but I will say take review once from @vvaibhavdesai

Issue #504, opened 2023-04-19
Tests for Taskdetails Page


kotesh-arya:
**Unit-Testing** for TaskDetails Page

- PR link https://github.com/Real-Dev-Squad/website-status/pull/341

PR #522, opened 2023-04-28
Migrate Taskdetails API to RTK Query

kotesh-arya:
- Added a custom hook using RTK query which is a replacement for the API calls of TaskDetails.
 - Added Tests for the custom hook.


src/interfaces/taskDetails.type.ts
```diff
@@ -0,0 +1,17 @@
+export type taskDetailsDataType = {
```
bhtibrewal, line 1: I think this should  be an interface

src/components/taskDetails/index.tsx
```diff
@@ -70,15 +70,15 @@ const TaskDetails: FC<Props> = ({ url, taskID }) => {
...
+    const { data, isError, isLoading } = useGetTaskDetailsQuery(taskID);
     const { SUCCESS, ERROR } = ToastTypes;
     const { taskDetails } = state;
     useEffect(() => {
-        const fetchedData: task = { ...response.taskData };
+        const fetchedData = data?.taskData;
         dispatch({ type: 'setTaskDetails', payload: fetchedData });
         initialDataRef.current = fetchedData;
-    }, [isLoading, response]);
+    }, [isLoading, data]);
```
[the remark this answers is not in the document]
kotesh-arya: In this case, the Superuser clicks `cancel` button, then I am resetting the title to initial-state by accessing the initial-data, which I am storing.

kotesh-arya:
> I don't see the task details page test that was failing being fixed here, are you doing that in a follow-up PR?

Yes, I have created a dedicated ticket for that and will address the tests in a follow-up PR
Issue Ticket: https://github.com/Real-Dev-Squad/website-status/issues/504

src/interfaces/taskDetails.type.ts
kotesh-arya: As I am **NOT EXTENDING** this by declaring it a second time anywhere, I have used the `type` here.

src/app/services/api.ts
```diff
@@ -14,12 +14,19 @@ export const api = createApi({
...
         'Tasks',
+        'TaskDetails',
+
         'Mine_Tasks',
         'TaskTag',
+
         'User',
+
         'Tags',
+
```
Muralidhar22, line 29: why additional space is added?

kotesh-arya: Removed them 👍🏻

src/components/taskDetails/index.tsx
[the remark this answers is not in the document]
kotesh-arya: **Update:** Migrated the API calls to RTK-Query completely, not using any dispatch now 👍🏻

src/app/services/api.ts
```diff
@@ -16,6 +16,7 @@ export const api = createApi({
         'Users',
         'Status',
         'Tasks',
+        'TaskDetails',
         'Mine_Tasks',
```
bhtibrewal, line 20: Let's follow the same naming convention

bhtibrewal (commented):
it looks okay to me, please address the comment

src/app/services/api.ts
kotesh-arya: Done 👍🏻

src/components/taskDetails/index.tsx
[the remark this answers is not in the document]
kotesh-arya: > This was not what I was trying pointing out, here I meant why do we need to store the fetchedData in another state here?

When the superuser starts editing the task-title/task-description and If incase they don't want to save it. Then they will click on the `cancel` and then the **initially** fetched data should be shown as the task-title and task-description.
In order to access that initially fetched data, I am storing the fetchedData in another state. Which will be shown when they cancel updating.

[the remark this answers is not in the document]
bhtibrewal: Kotesh RTK not only fetches data but stores it. for that initial data, you could have directly used that data.

bhtibrewal (approved):

shubhamsigdar1 (approved):
Great Work @kotesh-arya

sahsisunny (approved):

PR #547, opened 2023-05-11
Tests/dashboard page

RitikJaiswal75:
Add tests for standup dashboard page and search bar component

Test Coverage:
image


__tests__/Unit/Components/Searchbar/Searchbar.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+        const label = screen.getByLabelText('test-label');
+        expect(label).toBeInTheDocument();
+    });
+
+    it('checks if we can type into the searchbar', function () {
+        render(<Searchbar label="test-label" handleSearch={mockFn} />);
+
+        const input = screen.getByLabelText('test-label') as HTMLInputElement;
+
+        fireEvent.click(input);
```
bhtibrewal, line 23: why do we need to click on input?

__tests__/Unit/Components/Searchbar/Searchbar.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+        fireEvent.change(input, { target: { value: '123,456' } });
+        expect(input.value).toBe('123,456');
+    });
+
+    it('tests if the click handler is called', function () {
+        render(<Searchbar label="test-label" handleSearch={mockFn} />);
+
+        const input = screen.getByLabelText('test-label') as HTMLInputElement;
+
+        const searchButton = screen.getByRole('button');
```
bhtibrewal, line 33: what if in future we have, more buttons in the Search bar component, we should ensure this test will not fail because of that

bhtibrewal (commented):
Good work on the tests, left 2 small comments

__tests__/Unit/Components/Searchbar/Searchbar.test.tsx
RitikJaiswal75: simulatind user enent to click and type

__tests__/Unit/Components/Searchbar/Searchbar.test.tsx
RitikJaiswal75: I think there won't be more buttons rather we will write efficient queries, like we do in github

__tests__/Unit/Components/Searchbar/Searchbar.test.tsx
RitikJaiswal75: Oops! I missed this one, removed the click event.

__tests__/Unit/Components/Searchbar/Searchbar.test.tsx
RitikJaiswal75: and this search bar is specific to standup dashboard

vvaibhavdesai (approved):

prakashchoudhary07 (changes requested):

src/pages/dashboard/index.tsx
```diff
@@ -4,11 +4,6 @@ import PageNotFound from '../404';
 import NavBar from '@/components/navBar';
 import Searchbar from '@/components/Dashboard/Searchbar';
 
-const search = (query: string) => {
-    const searchValues = query.split(',');
-    console.log('Searching', searchValues);
-};
-
 const DashboardPage = () => {
```
iamitprakash, line 7: The dev property in the router's query is only set to true when the application is running in development mode. In production mode, the dev property will be undefined. This means that the if statement will always evaluate to false, and the PageNotFound component will always be rendered.

iamitprakash (changes requested):

src/pages/dashboard/index.tsx
iamitprakash: i want something like below

const DashboardPage = () => {
    const router = useRouter();

    const { dev } = router.query;
    if (typeof dev === "undefined") {
        return <PageNotFound />;
    }
    return (
        <>
            <NavBar />
            <Searchbar label="Users" />
        </>
    );
};

iamitprakash (changes requested):

src/pages/dashboard/index.tsx
RitikJaiswal75: it's not based on the URL being dev.realdevsquad.com

it is the query param that we pass in the URL. something like this `/?dev=true`

__tests__/Unit/pages/Dashboard/dashboardPage.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+        );
+
+        const searchBar = screen.getByRole('textbox');
+        expect(searchBar).toBeInTheDocument();
+
+        const searchButton = screen.getByRole('button');
+        expect(searchButton.innerHTML).toBe('Search');
+
+        const label = screen.getByLabelText('Users');
+        expect(label).toBeInTheDocument();
```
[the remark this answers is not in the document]
bharati-21: For this you can use screen.getByRole('button', { name: 'Search'}); This would get the specific element with role button and with text "Search".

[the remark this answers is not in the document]
bharati-21: Ref: https://testing-library.com/docs/queries/byrole/#api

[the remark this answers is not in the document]
bharati-21: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text

PR #563, opened 2023-05-16
Add RTK query and make task progress form work

RitikJaiswal75:
Created a progresses API in `/src/app/services` folder
Created a `mutation query`
it would call up the progress API and post the progress data

This PR closes: #564 

Tests for the feature are coming in consecutive PRs

[Screencast from 27-05-23 12:58:20 AM IST.webm](https://github.com/Real-Dev-Squad/website-status/assets/57758447/167c3584-e3b9-4eed-ad1e-f108e79eb789)

### DIFF
--- a/__tests__/Unit/Components/ProgressForm/ProgressForm.test.tsx
+++ b/__tests__/Unit/Components/ProgressForm/ProgressForm.test.tsx
@@ -1,9 +1,11 @@
 import { fireEvent, screen } from '@testing-library/react';
 
-import { renderWithProviders } from '@/test-utils/renderWithProvider';
+import { renderWithRouter } from '@/test_utils/createMockRouter';
 
 import ProgressForm from '@/components/ProgressForm/ProgressForm';
 import { questions } from '@/constants/ProgressUpdates';
+import { Provider } from 'react-redux';
+import { store } from '@/app/store';
 
 const mockOnClick = jest.fn();
 const mockQuestion = [
@@ -16,7 +18,11 @@ const mockQuestion = [
 
 describe('Progress form', function () {
     it('Should Render 3 input fields with appropriate data', function () {
-        renderWithProviders(<ProgressForm questions={questions} />);
+        renderWithRouter(
+            <Provider store={store()}>
+                <ProgressForm questions={questions} />
+            </Provider>
+        );
 
         const textBoxes = screen.getAllByRole('textbox');
         expect(textBoxes).toHaveLength(3);
@@ -26,7 +32,11 @@ describe('Progress form', function () {
     });
 
     it('Should change input values separately', function () {
-        renderWithProviders(<ProgressForm questions={questions} />);
+        renderWithRouter(
+            <Provider store={store()}>
+                <ProgressForm questions={questions} />
+            </Provider>
+        );
 
         const textAreas = screen.getAllByRole(
             'textbox'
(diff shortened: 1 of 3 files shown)


vvaibhavdesai (dismissed):

__tests__/Unit/Components/ProgressForm/ProgressForm.test.tsx
```diff
@@ -16,7 +18,11 @@ const mockQuestion = [
 
 describe('Progress form', function () {
```
iamitprakash, line 19: line renderWithRouter(.... This line is calling the renderWithRouter function, which renders the component in the context of a router. However, the ProgressForm component does not need to be rendered in the context of a router. Therefore, you can remove the renderWithRouter function call.

src/app/services/progressesApi.ts
```diff
@@ -0,0 +1,19 @@
+import { api } from './api';
+
+export const progressesApi = api.injectEndpoints({
```
iamitprakash, line 3: In line headers: { credentials: 'include' }. This line is setting the credentials header to include, which tells the browser to include the credentials from the current request when making the POST request to /progresses. However, the credentials header is only necessary when making a cross-origin request. Since the /progresses endpoint is on the same origin as the current request, there is no need to set the credentials header. Therefore, you can remove the headers: { credentials: 'include' }

src/app/services/progressesApi.ts
```diff
@@ -0,0 +1,19 @@
...
+    endpoints: (builder) => ({
```
iamitprakash, line 4: headers object is missing the Authorization header, which is required for authentication.

src/components/ProgressForm/ProgressForm.tsx
```diff
@@ -29,13 +35,43 @@ const reducer = (state: progressStates, action: reducerAction) => {
...
     const buttonSyle =
         state.progress && state.plan && state.blockers
             ? styles.buttonEnabled
             : styles.buttonDisabled;
 
+    if (isLoading) {
+        return <Loader></Loader>;
+    }
+
+    const handleSubmit = (e: MouseEvent) => {
```
iamitprakash, line 53: saveProgress function is asynchronous, so you need to use async and await to wait for the response before you can set the isLoading state to false

iamitprakash (changes requested):

__tests__/Unit/Components/ProgressForm/ProgressForm.test.tsx
RitikJaiswal75: Since i am getting the task id from route parameter so I am calling on useRouter in progress Form component, thus I have rendered it with router

src/app/services/progressesApi.ts
RitikJaiswal75: The cookie we set is a HTTP Only cookie, accessing it directly in frontend and setting it in headers is not possible thus credentials include will do the work for us

src/app/services/progressesApi.ts
RitikJaiswal75: credentials include is the one which sets it

src/components/ProgressForm/ProgressForm.tsx
RitikJaiswal75: since I am using `.then` thus this part can be skipped

iamitprakash (approved):

heyrandhir (approved):
LGTM

PR #575, opened 2023-05-20
Adding User-name suggestions to AssignedTo field

### LINKED ISSUE #560: Implement user search in the `Assigned to` Input
Currently, we write the name of the user manually

### Functionality

when we type in the input it should show a dropdown of the matching users

image

fakhruddinkw:
Closes #560 

To be done:

- [x] Make input field visible to type in name
- [x] On typing, there should be a suggestion box below input, matching the current input value

ETA - 25/05/2023

https://github.com/Real-Dev-Squad/website-status/assets/85249044/4d754e2c-6817-417b-b6f5-91d15e18e802

no-user-found

**RESPONSIVENESS**

https://github.com/Real-Dev-Squad/website-status/assets/85249044/0cc2716e-d87b-4c5f-a5c7-22d755bbe1c8

**EDIT WILL BE ONLY AVAILABLE IF EDIT=TRUE AND USER=SUPER_USER**

https://github.com/Real-Dev-Squad/website-status/assets/85249044/7b896703-6347-4db2-9ab6-c659eed0544a

https://github.com/Real-Dev-Squad/website-status/assets/85249044/093382c9-ed70-456c-a2d9-e22134c66190

It takes time for changes to reflect as you can see in the second video. But the changes are made successfully, as indicated by the success message in the first video


src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,107 @@
...
+type Props = {
+    assigneeName: string;
+    showSuggestion: boolean;
+    onClickName: (userName: string) => void;
+};
+
+type GithubInfo = {
+    github_id: string;
+    url: string;
+};
```
Pratiyushkumar, line 16: don't you think these types should be included in a separate files and exported into suggestionBox.tsx and then used

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,107 @@
...
+    showSuggestion = false,
+    onClickName = () => undefined,
+}) => {
+    const [loading, setLoading] = useState<boolean>(false);
+    const [suggestions, setSuggestions] = useState<GithubInfo[]>([]);
+
+    const clickHandler = (e: any) => {
+        onClickName(e.target.innerText);
+    };
+
```
Pratiyushkumar, line 29: you cannot use e of type any, if we need to use any then what is the use of typescript, we could have just used JS

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,107 @@
...
+        const usersData = users.data.users;
+        const suggestedUsers: GithubInfo[] = [];
+        usersData.map((data: userDataType) => {
+            suggestedUsers.push({
+                github_id: data.github_id,
+                url: data.picture
+                    ? data.picture.url
+                        ? data.picture.url
+                        : placeholderImageURL
+                    : placeholderImageURL,
```
Pratiyushkumar, line 44: please try to optimize this, it's confusing me

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,107 @@
...
+            });
+        });
+
+        setSuggestions(suggestedUsers);
+        setLoading(false);
+    };
+
+    useEffect(() => {
+        let timerId: any;
+        if (assigneeName) {
```
Pratiyushkumar, line 54: dont use type any, please, give it a sensible type

src/components/tasks/card/index.tsx
```diff
@@ -369,6 +375,16 @@ const Card: FC<Props> = ({
         );
     };
 
+    const handleAssigneeName = (e: any) => {
+        setAssigneeName(e.target.value);
+        setShowSuggestion(true);
+    };
```
Pratiyushkumar, line 381: use proper type please, stop using any

Pratiyushkumar (changes requested):
Also write test's

src/components/tasks/card/index.tsx
fakhruddinkw: @Pratiyushkumar sorry for using any.
This is my first time working with typescript, i did try with different types to make this work, but i was not able to so i used type any.
I will rectify this

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: Sure i wll refactor this.
In first condition we are just checking for if picture property exists or not. And in the nested condition we check if url property exists or not on picture object.
In both cases if it gives false, we use placeholderImageURL

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: I had thought of that, but then i assumed i only need this particular type only for this particular component so i thought why not localize it.

Pratiyushkumar: please make a different file for it and then import those types in the file and use it that would make your component look nice and optimized.

fakhruddinkw: Ok

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+import classes from '@/components/tasks/card/suggestion.module.scss';
+import fetch from '@/helperFunctions/fetch';
+import { userDataType } from '@/interfaces/user.type';
+import { GithubInfo } from '@/interfaces/suggestionBox.type';
+
+type Props = {
+    assigneeName: string;
+    showSuggestion: boolean;
+    onClickName: (userName: string) => void;
+};
```
Pratiyushkumar, line 12: please remove these also in suggestionBox.type

src/components/tasks/card/suggestion.module.scss
```diff
@@ -0,0 +1,64 @@
...
+    min-width: 7rem;
+    overflow: auto;
+    background-color: white;
+    position: absolute;
+    top: 20px;
+    left: -120px;
+    border-radius: 5px;
+    z-index: 100;
+    li {
+        padding: 0.5rem;
```
Pratiyushkumar, line 17: try to use one type of unit, if using rem please stick to rem wherever possible

src/components/tasks/card/suggestion.module.scss
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.no-suggestions {
+    color: #666464;
+    position: absolute;
+    top: 20px;
+    left: -120px;
+    z-index: 100;
+    min-width: 5rem;
+}
```
Pratiyushkumar, line 55: same as above

src/components/tasks/card/suggestion.module.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.loading {
+    color: #666464;
+    position: absolute;
+    top: 20px;
+    left: -120px;
+    z-index: 100;
```
Pratiyushkumar, line 62: here also

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+                        );
+                    })}
+                </ul>
+            );
+        } else {
+            suggestionList = (
+                <span className={classes['no-suggestions']}>
+                    User not found!
+                </span>
+            );
```
Pratiyushkumar, line 91: why does classname contain classes, please find some fruitful word for it, can confuse other dev's

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+            suggestedUsers.push({
+                github_id: data.github_id,
+                profileImageUrl: data.picture
+                    ? data.picture.url
+                        ? data.picture.url
+                        : placeholderImageURL
+                    : placeholderImageURL,
+            });
+        });
+
```
Pratiyushkumar, line 43: please optimize this, it isn't optimized

Pratiyushkumar (changes requested):
please try to address all the changes suggested you in the comment box and if fixed then reply that it is fixed.

src/components/tasks/card/suggestion.module.scss
fakhruddinkw: rem, as i believe will scale up or down depending on user's device. So it makes sense using it for margin, padding or size.
But here i do want this box to appear 20px from top and -120px from left, relative to the input box, regardless of user's device. So therefore I used an absolute unit like px. Please correct me, if my thinking was wrong.

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: Sorry, i missed this.

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: I believe 'classes' is self explaining here as to what it represents.

src/components/tasks/card/index.tsx
fakhruddinkw: fixed @Pratiyushkumar

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: fixed @Pratiyushkumar

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: I do not think we need to put prop types in different file? Props are localized, they belong to particular component.

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: fixed @Pratiyushkumar

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: resolved @Pratiyushkumar

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: fixed @Pratiyushkumar

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: fixed @Pratiyushkumar

src/components/tasks/card/SuggestionBox.tsx
vvaibhavdesai: follow either of the conventions check throughout the project and see how the types are actually structured if all types are taken out into a types file do it else if local types are kept local follow that.

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+type Props = {
+    assigneeName: string;
+    showSuggestion: boolean;
+    onClickName: (userName: string) => void;
+};
+
+const SuggestionBox: FC<Props> = ({
+    assigneeName,
+    showSuggestion = false,
+    onClickName = () => undefined,
```
vvaibhavdesai, line 17: why returns undefined ?

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+}) => {
+    const [loading, setLoading] = useState<boolean>(false);
+    const [suggestions, setSuggestions] = useState<GithubInfo[]>([]);
+
+    const clickHandler = (e: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
+        onClickName(e.currentTarget.innerText);
+    };
+
+    const fetchUsers = async () => {
+        setLoading(true);
```
vvaibhavdesai, line 27: try catch kaha hai ?

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+            });
+        });
+
+        setSuggestions(suggestedUsers);
+        setLoading(false);
+    };
+
+    useEffect(() => {
+        let timerId: NodeJS.Timeout;
+        if (assigneeName) {
```
vvaibhavdesai, line 48: what are we trying to do here ?

src/components/tasks/card/suggestion.module.scss
vvaibhavdesai: does make sense to what you said also @fakhruddinkw please check on responsiveness that this doesnt break anything

fakhruddinkw: I have tested this on different screens, It works @vvaibhavdesai

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: So, i do not want to send request on every keystroke, so i just introduced a timeout, now if user is continuously typing I do not want to send the request, and hence i am storing the timerId, so that i can clean up that particular setTimeOut out of the queue. I hope this makes sense @vvaibhavdesai

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: It does not return anything but is just changing the state in the parent prop.

fakhruddinkw: and for the 'type Props' object, yes this is the convention we are following throughout the project.
@vvaibhavdesai

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: coming in the next push @vvaibhavdesai

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: for the most part, consisting of typeScript, I am just looking around the repo and figuring out. This is the first time i am working with it @vvaibhavdesai

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: resolved @Pratiyushkumar

src/components/tasks/card/suggestion.module.scss
bhtibrewal: you are correct on this @fakhruddinkw its not a rule to stick with one unit, we need to understand which to use where

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+            );
+        }
+    }
+
+    return (
+        <div style={{ position: 'relative' }}>
+            {loading && (
+                <span className={classNames['loading']}>Loading..</span>
+            )}
+            {!loading && suggestionList}
```
bhtibrewal, line 98:
```suggestion
           if(loading) return <Loader/>
           if (suggestion.length) return <SuggestionList/>
           return <UserNotFound />
```

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const [suggestions, setSuggestions] = useState<GithubInfo[]>([]);
+
+    const clickHandler = (e: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
+        onClickName(e.currentTarget.innerText);
+    };
+
+    const fetchUsers = async () => {
+        setLoading(true);
+        const url = `${process.env.NEXT_PUBLIC_BASE_URL}/users?search=${assigneeName}`;
+        const { requestPromise } = fetch({ url });
```
bhtibrewal, line 29: Please create a separate issue to migrate this fetch implementation to RTK Query, and link it here

fakhruddinkw: Ok. I will do it

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: ok ok you want to me to break this even further into smaller pieces of code? @bhtibrewal

fakhruddinkw: I did try what you asked, but the problem is if do this, The 'User not found!' persists, even if we have no value in the input.
I think that is bad experience. I want 'User not found' only when there is indeed no user found for the particular input value.

shubhamsigdar1:
Hi @fakhruddinkw Can you please add before and after image/video in description
For more visualization

src/components/tasks/card/index.tsx
```diff
@@ -392,7 +408,7 @@ const Card: FC<Props> = ({
                         <span
                             className={classNames.cardTitle}
                             contentEditable={shouldEdit}
-                            onKeyPress={(e) => handleChange(e, 'title')}
+                            onKeyDown={(e) => handleChange(e, 'title')}
```
shubhamsigdar1, line 411: Hey @fakhruddinkw why you change from 
onKeyPress to onKeyDown here

fakhruddinkw: onKeyPress is deprecated. Its use is not recommended. And onKeyDown does the same job @shubhamsigdar1

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: Okay so I was able to work around that. But with a minor issue: We still render the SuggestionBox element, but ,in case there is not input it renders empty react fragments. @bhtibrewal

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: @vvaibhavdesai so the reason i did not add try catch was because i tested the conditions in which it may cause the error.
So the possible error causing scenario would be, if we are querying with an empty input value, in such case we get the error, that 'search input cannot be empty'. Now what i did was introduced 'if check on assigneeName', now if there is some value in assigneeName only then we make the request.
Now what i did was obviously a gross assumption, because there can be plenty of other reason for the request to return an error. so yes i did add a try catch in the fetch

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: @bhtibrewal never mind figured out a workaround the problem, it is fixed

src/components/tasks/card/SuggestionBox.tsx
vvaibhavdesai: use ()=>void instead of undefined

src/components/tasks/card/SuggestionBox.tsx
vvaibhavdesai: I think a better way to do this would be using useCallback rather than using a useEffect

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,121 @@
...
+        return (
+            <span className={classNames['no-suggestions']}>
+                User not found!
+            </span>
+        );
+    };
+
+    if (loading)
+        return (
+            <div style={{ position: 'relative' }}>
```
vvaibhavdesai, line 104: your all three states have ?
 <div style={{ position: 'relative' }}>
 this in common 
 so rather make a variable out something like
 
``` js
let renderComponent;

if (loading){
renderComponent = <Loading/>
} .... 



return  <div style={{ position: 'relative' }}>{renderComponent}</div>

File: src/components/tasks/card/card.module.scss [hunk 19]
```diff
@@ -103,6 +103,10 @@ $seeMoreTasksShadow: #595959cc;
 .cardStrongFont {
     font-size: 1.3rem;
     font-weight: 500;
+    border-radius: 5px;
+    background-color: #adadad;
```
vvaibhavdesai, line 107: use css variables instead

src/components/tasks/card/SuggestionBox.tsx
fakhruddinkw: Can you please explain on the useCallBack part? 
I am confused bcz i thought useCallback is used to memoize the function? @vvaibhavdesai

fakhruddinkw: @vvaibhavdesai void is giving me syntax error

src/components/tasks/card/SuggestionBox.tsx
vvaibhavdesai: image
you dont need to do this 
just write `onClickName`

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    const Loading = () => {
+        return <span className={classNames['loading']}>Loading...</span>;
+    };
+
+    const UserNotFound = () => {
+        return (
+            <span className={classNames['no-suggestions']}>
+                User not found!
```
bhtibrewal, line 97: These components should be independent and outside the parent component

src/components/tasks/card/SuggestionBox.tsx
```diff
@@ -0,0 +1,112 @@
...
+        );
+    };
+
+    const Loading = () => {
+        return <span className={classNames['loading']}>Loading...</span>;
+    };
+
+    const UserNotFound = () => {
+        return (
+            <span className={classNames['no-suggestions']}>
```
shubhamsigdar1, line 96: Nit: @fakhruddinkw just asking for learning purpose...why you write className={classNames['no-suggestions']}
not className = 'no-suggestions' ??

shubhamsigdar1 (commented):
great work @fakhruddinkw and Thanku for adding in description the video

src/components/tasks/SuggestionBox/SuggestionList.tsx
```diff
@@ -0,0 +1,38 @@
+import { FC } from 'react';
+import { GithubInfo } from '@/interfaces/suggestionBox.type';
+import classNames from '@/components/tasks/SuggestionBox/suggestion.module.scss';
+
+type Props = {
+    suggestions: GithubInfo[];
+    onClickName: (userName: string) => void;
+};
+
+const SuggestionList: FC<Props> = ({ suggestions, onClickName }) => {
```
iamitprakash, line 10: onClickName function is expecting a string as its argument, but the innerText property of an HTML element returns a DOMText node. This means that the onClickName function will receive a DOMText node instead of a string, which will cause an error.

src/components/tasks/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const [loading, setLoading] = useState<boolean>(false);
+    const [suggestions, setSuggestions] = useState<GithubInfo[]>([]);
+
+    const { ERROR } = ToastTypes;
+
+    const fetchUsers = async () => {
+        setLoading(true);
+        const url = `${process.env.NEXT_PUBLIC_BASE_URL}/users?search=${assigneeName}`;
+        try {
+            const { requestPromise } = fetch({ url });
```
RitikJaiswal75, line 26: is it a good call to use fetch in a component?
Please use rtk query as the whole status site is moving towards it.

iamitprakash (changes requested):

src/components/tasks/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,78 @@
...
+            setLoading(false);
+            if ('response' in error) {
+                toast(ERROR, error.response.data.message);
+                return;
+            }
+            toast(ERROR, error.message);
+        }
+    };
+
+    useEffect(() => {
```
iamitprakash, line 51: useEffect hook is not being called when the assigneeName prop changes. This means that the fetchUsers function will only be called once, when the component is first mounted.

iamitprakash (changes requested):

src/components/tasks/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,26 @@
+import React, { FC } from 'react';
```
Pratiyushkumar, line 1: please remove this react import if you are not using this react

src/components/tasks/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { GithubInfo } from '@/interfaces/suggestionBox.type';
+import SuggestionList from './SuggestionList';
+import Loading from './Loading';
+import UserNotFound from './UserNotFound';
+
+type Props = {
+    onClickName: (userName: string) => void;
+    loading: boolean;
+    suggestions: GithubInfo[];
+};
```
Pratiyushkumar, line 11: can we please export these types from .type file

src/components/tasks/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const SuggestionBox: FC<Props> = ({ onClickName, suggestions, loading }) => {
+    let renderComponent = (
+        <SuggestionList suggestions={suggestions} onClickName={onClickName} />
+    );
+
+    if (loading) renderComponent = <Loading />;
+    else if (suggestions.length === 0) {
+        renderComponent = <UserNotFound />;
```
Pratiyushkumar, line 20: please use blocks in if loop to make it readable

src/components/tasks/SuggestionBox/SuggestionList.tsx
```diff
@@ -0,0 +1,38 @@
+import { FC } from 'react';
+import { GithubInfo } from '@/interfaces/suggestionBox.type';
+import classNames from '@/components/tasks/SuggestionBox/suggestion.module.scss';
+
```
Pratiyushkumar, line 4: please  use another name instead of classNames as it is an HTML attribute in React

src/components/tasks/SuggestionBox/SuggestionList.tsx
```diff
@@ -0,0 +1,38 @@
...
+type Props = {
+    suggestions: GithubInfo[];
+    onClickName: (userName: string) => void;
+};
```
Pratiyushkumar, line 8: please export types from .type file, focus on making your code look clean

src/components/tasks/SuggestionBox/suggestion.module.scss
Pratiyushkumar, line 1: please follow the design guidelines and prefer one unit either px or em or rem

src/components/tasks/card/card.module.scss
```diff
@@ -103,6 +104,10 @@ $seeMoreTasksShadow: #595959cc;
 .cardStrongFont {
     font-size: 1.3rem;
     font-weight: 500;
+    border-radius: 5px;
+    background-color: $grey;
+    width: 7rem;
+    height: 1.5rem;
 }
```
Pratiyushkumar, line 112: same as above follow the design guidelines and stick to rem or em or px

src/components/tasks/card/index.tsx
```diff
@@ -31,6 +31,9 @@ import { useEditMode } from '@/hooks/useEditMode';
 import { useGetUsersByUsernameQuery } from '@/app/services/usersApi';
 import { ConditionalLinkWrapper } from './ConditionalLinkWrapper';
 import { isNewCardDesignEnabled } from '@/constants/FeatureFlags';
+import SuggestionBox from '../SuggestionBox/SuggestionBox';
+import { userDataType } from '@/interfaces/user.type';
+import { GithubInfo } from '@/interfaces/suggestionBox.type';
 
 type Props = {
     content: task;
```
Pratiyushkumar, line 39: please import types from .type file and use it here

src/components/tasks/card/index.tsx
```diff
@@ -85,6 +98,8 @@ const Card: FC<Props> = ({
         }
     }, [keyLongPressed]);
 
+    const [showSuggestion, setShowSuggestion] = useState(false);
+
```
Pratiyushkumar, line 102: mention here type of  show suggestion  as boolean

Pratiyushkumar (changes requested):
Please make all the changes and also see if your UI is responsive or not, please record the video of responsiveness and put it in description

src/components/tasks/card/index.tsx
fakhruddinkw: This is the way, it has been done across the whole repo. So i went along with it @Pratiyushkumar

src/components/tasks/SuggestionBox/SuggestionList.tsx
fakhruddinkw: commented below @Pratiyushkumar

src/components/tasks/card/card.module.scss
fakhruddinkw: Again, i think it is not necessary to stick to one unit, we have to consider if it needs scaling up or down. I do not want the border-radius to scale up/down that is why it is in absolute unit like px.

src/components/tasks/SuggestionBox/SuggestionList.tsx
fakhruddinkw: This is the name used across the whole repo for import classes from modules. @Pratiyushkumar

src/components/tasks/SuggestionBox/SuggestionBox.tsx
fakhruddinkw: I have commented down below regarding this @Pratiyushkumar

Pratiyushkumar (approved):
Looks good to me

vvaibhavdesai (dismissed):

iamitprakash (approved):
Done

shubhamsigdar1:
Hello @fakhruddinkw, I get confused after seeing the video
When super_user press **Alt** then he can type the name on input field

and after typing the name the input field disappear and we can see like before
i.e. like this image
image

src/components/tasks/card/index.tsx
```diff
@@ -574,15 +652,38 @@ const Card: FC<Props> = ({
...
-                                className={classNames.cardStrongFont}
-                                contentEditable={shouldEdit}
-                                onKeyPress={(e) => handleChange(e, 'assignee')}
-                                role="button"
-                                tabIndex={0}
+                            <div
+                                style={{
+                                    position: 'relative',
+                                    display: 'inline-block',
+                                }}
```
shubhamsigdar1, line 659: inline styling :)

shubhamsigdar1 (commented):
Hi @fakhruddinkw great work...I written some comment can you please reply to those comment
and also where you written test??

src/components/tasks/card/index.tsx
fakhruddinkw: Yes. Inline styling :) @shubhamsigdar1

fakhruddinkw:
> Hi @fakhruddinkw great work...I written some comment can you please reply to those comment and also where you written test??

Shubham tests are ready to go. I just need this PR to be merge, so i can push it, as this PR got bigger i am keeping it separate.

vvaibhavdesai (approved):

Pratiyushkumar (approved):

src/components/tasks/SuggestionBox/SuggestionBox.tsx
```diff
@@ -0,0 +1,27 @@
...
+        <SuggestionList suggestions={suggestions} onClickName={onClickName} />
+    );
+
+    if (loading) {
+        renderComponent = <Loading />;
+    } else if (suggestions.length === 0) {
+        renderComponent = <UserNotFound />;
+    }
+
+    return <>{renderComponent}</>;
```
bhtibrewal, line 24: nitpik: 
```suggestion


  if (loading) 
        return <Loading />;
 if (suggestions.length === 0) 
        return <UserNotFound />;
    
    return  <SuggestionList suggestions={suggestions} onClickName={onClickName} />
```

src/components/tasks/card/index.tsx
```diff
@@ -62,8 +68,11 @@ const Card: FC<Props> = ({
     const isUserAuthorized = useContext(isUserAuthorizedContext);
 
     const [showEditButton, setShowEditButton] = useState(false);
+
     const [keyLongPressed] = useKeyLongPressed();
 
+    // const [assigneeName, setAssigneeName] = useState('');
```
bhtibrewal, line 74: please remove the commented code

src/components/tasks/card/index.tsx
```diff
@@ -442,15 +509,46 @@ const Card: FC<Props> = ({
...
+                                    />
+
+                                    {showSuggestion && (
+                                        <SuggestionBox
+                                            suggestions={suggestions}
+                                            onClickName={handleClick}
+                                            loading={isLoadingSuggestions}
+                                        />
+                                    )}
+                                </div>
```
bhtibrewal, line 545: Suggestion: we can take this out into a separate component it will improve the code readability

bhtibrewal (commented):
Left few small suggestions, please look into them, in okay you can address them in a follow-up

bhtibrewal (approved):

fakhruddinkw:
> Left few small suggestions, please look into them, in okay you can address them in a follow-up

Thank you @bhtibrewal ! I will clean this up when i push the tests in next PR. Thank you again for understanding 
🙏

shubhamsigdar1 (approved):
Great work @fakhruddinkw

iamitprakash (approved):

src/components/tasks/SuggestionBox/SuggestionBox.tsx
RitikJaiswal75, line 24: Could we remove the fragment here the renderComponent does not contain multiple tags right? fragment is used when we want to return multiple HTML elements/tags to consolidate them into one.

src/components/tasks/card/index.tsx
RitikJaiswal75: Please remove the commented code

src/components/tasks/card/index.tsx
```diff
@@ -131,6 +148,7 @@ const Card: FC<Props> = ({
                     new Date(`${event.target.value}`).getTime() / 1000;
                 toChange[changedProperty] = toTimeStamp;
             }
+            console.log(toChange);
```
RitikJaiswal75, line 151: Could we remove this console log? I don't think we need console log when the code runs in production.

src/components/tasks/card/index.tsx
```diff
@@ -369,6 +387,55 @@ const Card: FC<Props> = ({
         );
     };
 
+    const handleAssignment = (e: React.ChangeEvent<HTMLInputElement>) => {
+        setAssigneeName(e.target.value);
+        if (e.target.value) setShowSuggestion(true);
+        else setShowSuggestion(false);
```
RitikJaiswal75, line 393: nit: can we use the ternary operator `()? doSomething() : doSomething();` or curly braces `{}`

src/components/tasks/card/index.tsx
```diff
@@ -442,15 +509,46 @@ const Card: FC<Props> = ({
...
-                        >
-                            {cardDetails.assignee}
-                        </span>
+                        {shouldEdit ? (
+                            isUserAuthorized && (
+                                <div
+                                    style={{
+                                        position: 'relative',
+                                        display: 'inline-block',
+                                    }}
```
RitikJaiswal75, line 518: can we use class instead of inline style?

src/components/tasks/card/index.tsx
```diff
@@ -574,15 +672,48 @@ const Card: FC<Props> = ({
...
-                            >
-                                {cardDetails.assignee}
-                            </span>
+                            {shouldEdit ? (
+                                isUserAuthorized && (
+                                    <div
+                                        style={{
+                                            position: 'relative',
+                                            display: 'inline-block',
+                                        }}
```
RitikJaiswal75, line 681: use classes for styling please

src/interfaces/suggestionBox.type.ts
RitikJaiswal75, line 1: hey @fakhruddinkw the types now go in a folder named types please use that, soon enough all the types files in interfaces will be moved to the types folder as well.

RitikJaiswal75 (changes requested):

src/components/tasks/SuggestionBox/Loading.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+const Loading = () => {
+    return (
+        <div className={classNames['loading']} data-testid="loading">
+            Loading...
+        </div>
+    );
+};
+
+export default Loading;
```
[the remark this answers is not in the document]
fakhruddinkw: The loader renders a spinner, but i didnt want that. I want a Loading... text. So that is why i created a new component @RitikJaiswal75

src/components/tasks/card/index.tsx
fakhruddinkw: This were highly localized style so i went for in-line styles

src/interfaces/suggestionBox.type.ts
fakhruddinkw: I have moved the types to the interfaces folder already. Did i miss something? @RitikJaiswal75

RitikJaiswal75: Um, the expectation was to move interfaces to the types folder sorry if I wasn't clear here

RitikJaiswal75 (approved):

PR #584, opened 2023-05-26
fix: Showing User as Logged in even is Logged Out

WYGIN:
### Description:
When the user is not logged in the website is showing similar screen with profile, similar to when the user is loggedin

### Dev Tested:
- [x] Yes

#### After: 
Screenshot_20230526-163742


shubhamsigdar1:
add before and after image in description

bhtibrewal (approved):
looks okay to me

PR #593, opened 2023-05-28
add getLatestProgress RTK query to make missed updates dynamic

RitikJaiswal75:
### Issue:
Add RTK query to fetch the latest progress 

### Description:
The Page always showed you have 2 missed updates, Fixed it to show this dynamically based on API call

Tests are on the way in the consecutive PRs

### Dev Tested:
- [x] Yes

### Images/video of the change: (you can see the different values for missed updates in each image)
image
image

### Follow-up Issues (if any)
Integration Tests for the task progress Feature
https://github.com/Real-Dev-Squad/website-status/issues/594

### DIFF
--- a/src/app/services/progressesApi.ts
+++ b/src/app/services/progressesApi.ts
@@ -17,7 +17,8 @@ export const progressesApi = api.injectEndpoints({
         getLatestProgress: builder.query({
             query: (payload) => `/progresses/?${payload}`,
             transformResponse: (response: taskProgressResponse) => {
-                return response.data[0];
+                if (response.data.length > 0) return response.data[0];
+                else return { id: '', date: Date.now() };
             },
         }),
     }),
--- a/src/app/services/progressesApi.ts
+++ b/src/app/services/progressesApi.ts
@@ -17,7 +17,7 @@ export const progressesApi = api.injectEndpoints({
         getLatestProgress: builder.query({
             query: (payload) => `/progresses/?${payload}`,
             transformResponse: (response: taskProgressResponse) => {
-                return response.data[0].date;
+                return response.data[0];
             },
         }),
     }),
--- a/src/app/services/progressesApi.ts
+++ b/src/app/services/progressesApi.ts
@@ -1,3 +1,4 @@
+import { taskProgressResponse } from '@/types/ProgressUpdates';
 import { api } from './api';
 
 export const progressesApi = api.injectEndpoints({
@@ -13,7 +14,14 @@ export const progressesApi = api.injectEndpoints({
                 },
             }),
         }),
+        getLatestProgress: builder.query({
+            query: (payload) => `/progresses/?${payload}`,
+            transformResponse: (response: taskProgressResponse) => {
+                return response.data[0].date;
+            },
+        }),
     }),
 });
 
-export const { useSaveProgressMutation } = progressesApi;
+export const { useSaveProgressMutation, useGetLatestProgressQuery } =
+    progressesApi;
--- a/src/components/ProgressForm/ProgressHeader.tsx
+++ b/src/components/ProgressForm/ProgressHeader.tsx
@@ -5,18 +5,18 @@ import { useRouter } from 'next/router';
 function ProgressHeader() {
     const router = useRouter();
     const id = router.query.id;
-    const { data } = useGetLatestProgressQuery(`taskId=${id}`);
-    const currentDate = new Date();
-    const lastUpdateDate = new Date(data?.date as number);
+    const { data, error } = useGetLatestProgressQuery(`taskId=${id}`);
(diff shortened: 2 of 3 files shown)


heyrandhir (dismissed):

src/app/services/progressesApi.ts
```diff
@@ -13,7 +14,14 @@ export const progressesApi = api.injectEndpoints({
                 },
             }),
         }),
+        getLatestProgress: builder.query({
+            query: (payload) => `/progresses/?${payload}`,
+            transformResponse: (response: taskProgressResponse) => {
+                return response.data[0];
```
iamitprakash, line 20: what if data array is null

iamitprakash (changes requested):

shubhamsigdar1 (approved):

src/app/services/progressesApi.ts
RitikJaiswal75: thanks for pointing out, fixed this

src/components/ProgressForm/ProgressHeader.tsx
```diff
@@ -1,11 +1,23 @@
...
 
 function ProgressHeader() {
+    const router = useRouter();
+    const id = router.query.id;
+    const { data, error } = useGetLatestProgressQuery(`taskId=${id}`);
+    let missedUpdates = 0;
+    if (!error) {
+        const currentDate = new Date();
+        const lastUpdateDate = new Date(data?.date as number);
+        missedUpdates = currentDate.getDate() - 1 - lastUpdateDate.getDate();
```
prakashchoudhary07, line 13: How does this work?
Are we getting the difference between the 2 dates here?

prakashchoudhary07 (changes requested):

src/components/ProgressForm/ProgressHeader.tsx
TanishqSingla, line 13: We already have `moment` library for time, can this be solved by that?

TanishqSingla (changes requested):

src/app/services/progressesApi.ts
```diff
@@ -13,7 +14,15 @@ export const progressesApi = api.injectEndpoints({
                 },
             }),
         }),
+        getLatestProgress: builder.query({
+            query: (payload) => `/progresses/?${payload}`,
+            transformResponse: (response: taskProgressResponse) => {
```
iamitprakash, line 19: transformResponse function for the getLatestProgress endpoint is incorrectly returning an object with an empty id property if there are no results in the response

iamitprakash (changes requested):

src/types/ProgressUpdates.d.ts
```diff
@@ -30,3 +30,13 @@ interface question {
 export interface formProps {
     questions: Array<question>;
 }
+
+export interface fetchedProgress {
+    id: string;
+    date: number;
```
iamitprakash, line 36: why date is in number

iamitprakash (changes requested):

src/types/ProgressUpdates.d.ts
RitikJaiswal75: As per the DTO decided by the backend, the API call returns a number
Also, the date constructor uses a number as a parameter.

vvaibhavdesai (changes requested):

src/components/ProgressForm/ProgressHeader.tsx
RitikJaiswal75: yes we are

src/components/ProgressForm/ProgressHeader.tsx
```diff
@@ -1,11 +1,23 @@
+import { useGetLatestProgressQuery } from '@/app/services/progressesApi';
 import styles from '@/components/ProgressForm/ProgressForm.module.scss';
+import { useRouter } from 'next/router';
 
 function ProgressHeader() {
+    const router = useRouter();
+    const id = router.query.id;
+    const { data, error } = useGetLatestProgressQuery(`taskId=${id}`);
+    let missedUpdates = 0;
+    if (!error) {
```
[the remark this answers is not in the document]
RitikJaiswal75: then it will show 0 missed updates.

src/components/ProgressForm/ProgressHeader.tsx
prakashchoudhary07: So if `currentDate` is 1 and `lastUpdatedDate` is 31 then the missing updates will be 1-1-31 = -31?
Correct me if I am missing something here.

And why are we subtracting -1 from it?

prakashchoudhary07 (changes requested):

Issue #430, opened 2023-02-23
Tabs component


ankushdharkar:
# Status Tasks Tabs
## Overview

In Real Dev Squad, we use the [status site](https://status.realdevsquad.com/) to keep track of ongoing tasks. This a custom project tracking solution for our usecases.

It's time to refresh our designs and start making things a bit more consistent with the overall brand. So, we are starting off with updating or creating our components.

We want to implement this new design for tabs that help us filter the tasks list shown on the main page.

## Design
[Figma Link](https://www.figma.com/file/NzTU7vD56xTmIls5CoWT4a/Status-Tabs?node-id=0%3A1&t=Kj9vtWdEXzTnuwQf-1)

## User stories

- As a viewer, the tabs should match the design provided.
- As a viewer, the tabs should correctly render the information to match the current implementation.
- The correct count summary should be available for each kind of tab.

## Requirements

- TDD should be followed.
- Put the component usage behind a FF(feature flag).

## Acceptance Criteria

- [ ] The new component can seemlessly be swapped out without affecting the functionality of the existing page.
- [ ] Selecting the tab renders the correct list of tasks on the page.
- [ ] Tests coverage is exhaustive enough for the component to guarantee correctness and stability.

kedar-hargude:
I would like to work on this!

PR #598, opened 2023-05-30
feat: update Tabs component with count summary

kedar-hargude:
### Issue:
update the Tabs component with the count summary of each Tab

### Description:
The Tabs component doesn't show the summary count of the number of tasks for each tab. 
Updating the component, upon using feature flag, would show the correct count to the viewer, alongside the tab name and change the design of the tab according to the figma file.

### Dev Tested:
- [X] Yes

image

### Test Coverage
Screenshot from 2023-06-29 08-31-47
Screenshot from 2023-06-29 08-32-21

#430

### DIFF
--- a/src/components/Tabs/Tabs.module.scss
+++ b/src/components/Tabs/Tabs.module.scss
@@ -1,11 +1,8 @@
 $theme-grey: #d4d4d5;
-<<<<<<< HEAD
 $theme-grey-light: #f1f1f5;
 $gray-shade-color: #808080;
 $gray-shade-color2: #5c5b5b;
-=======
 $theme-dark-grey: #a1a5b7;
->>>>>>> e03a416 (refactor: create variable name for colours)
 $theme-pink: #e30062;
 
 .tabButton {


src/components/Tabs/Tabs.module.scss
```diff
@@ -10,9 +10,27 @@ $theme-grey: #d4d4d5;
     cursor: pointer;
 }
 
+.featureFlagTabButton {
+    background: white;
+    border: none;
+    padding: 12px 8px 8px;
+    font-size: 0.8em;
```
harshith-venkatesh, line 17: Could you check whether em suits better or rem here

src/components/Tabs/Tabs.module.scss
```diff
@@ -10,9 +10,27 @@ $theme-grey: #d4d4d5;
...
+    border: none;
+    cursor: pointer;
+    color: #a1a5b7;
```
harshith-venkatesh, line 20: Would it be better to create variable name for color?

kedar-hargude: I didn't use a variable name since it's being used only once. But if it's better, then I'll surely use it.

src/components/Tabs/Tabs.module.scss
kedar-hargude: Checked it for all dimensions. There is no visible change for such a short value.

RitikJaiswal75 (changes requested):
Please provide test coverage in PR description

PR #622, opened 2023-06-07
Change fetch implementation to RTK Query for Idle Members

Aryex82:
### Description:
currently are using Javscripts Fetch API for GET  /members/idle, this PR replaces it with RTK query.
### Dev Tested:
- [x] Yes

Screenshot 2023-06-09 at 9 27 32 PM


src/pages/availability-panel/index.tsx
```diff
@@ -36,30 +41,11 @@ const AvailabilityPanel: FC = () => {
...
         fetchTasks();
-        fetchIdleUsers();
     }, [refreshData]);
 
     let isErrorOrIsLoading;
-    if (error) {
+    if (error || membersError) {
         isErrorOrIsLoading = (
             <span className={classNames.statusMessage}>
                 Something went wrong, please contact admin!
```
Pratiyushkumar, line 51: if we have membersError then why are we using error

src/pages/availability-panel/index.tsx
```diff
@@ -9,14 +9,19 @@ import updateTasksStatus from '@/helperFunctions/updateTasksStatus';
...
     const [error, setError] = useState<boolean>(false);
     const [isTaskLoading, setIsTaskLoading] = useState<boolean>(true);
-    const [isMemberLoading, setIsMemberLoading] = useState<boolean>(true);
     const [refreshData, setRefreshData] = useState<boolean>(false);
+    const {
+        data: idleMembersList = [],
+        isError: membersError,
+        isLoading: isMemberLoading,
+        refetch: refreshMemberList,
+    } = useGetIdleMembersQuery();
```
Pratiyushkumar, line 24: if you are not using isLoading  or reFetch then there is no point to write it here

src/pages/availability-panel/index.tsx
```diff
@@ -9,14 +9,19 @@ import updateTasksStatus from '@/helperFunctions/updateTasksStatus';
 import { AVAILABLE } from '@/constants/task-status';
 import { FEATURE } from '@/constants/task-type';
 import { BASE_URL } from '@/constants/url';
+import { useGetIdleMembersQuery } from '@/app/services/membersApi';
 
 const AvailabilityPanel: FC = () => {
-    const [idleMembersList, setIdleMembersList] = useState<string[]>([]);
     const [unAssignedTasks, setUnAssignedTasks] = useState<task[]>([]);
     const [error, setError] = useState<boolean>(false);
     const [isTaskLoading, setIsTaskLoading] = useState<boolean>(true);
```
Pratiyushkumar, line 17: Also, see if these states are required or if can we use the error, loading property from the RTK query that you have written or already written RTK query.

Pratiyushkumar: try to minimize the use state hook if possible

Pratiyushkumar (changes requested):
Also please add the data and screen-recorded output to the PR description, you PR description does not give any information to the reviewer

src/pages/availability-panel/index.tsx
Aryex82: These are for a different fetch call.

src/pages/availability-panel/index.tsx
Aryex82: Both are used. `isMemberLoading ` was previously used as a state variable so its not reflected in this PR and `refreshMemberList ` is used in the `getData` method.

src/pages/availability-panel/index.tsx
Aryex82: As there are two different fetch calls happening, if we use same variable for both the errors I feel it will increase the code complexity unnecessarily.

Aryex82:
> Also please add the data and screen-recorded output to the PR description, you PR description does not give any information to the reviewer


As there was no change to UI in this PR I wasn't sure if a screenshot was necessary. I've added one now.

__tests__/Unit/hooks/membersApi.test.tsx
```diff
@@ -0,0 +1,44 @@
+import { setupServer } from 'msw/node';
+import handlers from '../../../__mocks__/handlers/members.handler.js';
+import React, { PropsWithChildren } from 'react';
+import { act, renderHook } from '@testing-library/react-hooks';
```
Pratiyushkumar, line 4: why are we importing react if we are not using it

src/pages/availability-panel/index.tsx
Pratiyushkumar: do we need fetch calls if we are using rtk query, do figure it out

src/pages/availability-panel/index.tsx
Pratiyushkumar: please read about what refetch does in rtk query then use it

src/pages/availability-panel/index.tsx
Pratiyushkumar: please see the solution in terms of rtk, we are removing fetch calls, in the next 2 or 3 PR's fetch calls will be removed so think in terms of rtk

Aryex82: Yes since each RTK query returns a separate error variable the logic will remain same in the future.

src/pages/availability-panel/index.tsx
Aryex82: From what I understand it makes a new request to fetch the current data from the server. Which I believe is same as what was happening in the previous fetch call.

src/pages/availability-panel/index.tsx
Aryex82: Ideally no, The Issue this PR solves only asks to migrate idle members fetch call. Would be happy to migrate the other fetch call in a different PR if no one else is working on it.

__tests__/Unit/hooks/membersApi.test.tsx
Aryex82: removed it.

Pratiyushkumar (approved):
LGTM

bhtibrewal (approved):
LGTM

src/pages/availability-panel/index.tsx
```diff
@@ -9,14 +9,19 @@ import updateTasksStatus from '@/helperFunctions/updateTasksStatus';
...
-    const [isMemberLoading, setIsMemberLoading] = useState<boolean>(true);
     const [refreshData, setRefreshData] = useState<boolean>(false);
+    const {
+        data: idleMembersList = [],
+        isError: membersError,
```
bhtibrewal, line 21: nitpik: the name `membersError` doesn't seem appropriate to describe the value in the variable something like `isIdleMembersError` would have been better

Pratiyushkumar (approved):
LGTM

__tests__/Unit/hooks/membersApi.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+        await act(() => waitForNextUpdate());
+
+        const nextResponse = result.current;
+        expect(nextResponse.data).not.toBeUndefined();
+        expect(nextResponse.isLoading).toBe(false);
+        expect(nextResponse.isSuccess).toBe(true);
+        expect(nextResponse.data).toBeDefined();
+    });
+});
```
bharati-21, line 44: Please also have tests for failure cases.

Aryex82: Since we are not sending any request data in this API, I can't think of any negative scenarios. Im not sure if failures like status 400 & 500 needs to be tested here. I feel checking status 400 & 500 makes more sense while testing a component.

bharati-21: Great work!

bharati-21 (approved):
LGTM

bharati-21 (changes requested):
Please resolve conflicts. Post that I'll approve.

bharati-21 (approved):
LGTM

Issue #526, opened 2023-05-04
Refactor Task Card Component to Improve Readability and Maintainability


sahsisunny:
**Description**
The[ src/components/tasks/card/index.tsx ](https://github.com/Real-Dev-Squad/website-status/blob/30225429984dcdcf746fdef59f92705ee4e15f82/src/components/tasks/card/index.tsx)file is becoming too complex with a lot of code stuffed into it. To improve the readability and maintainability of the code, let's refactor the Task Card component and break it down into smaller, more manageable pieces.

Some potential steps we could take to achieve this include:

- Identify logical sections of the component that can be split into separate sub-components.
- Extract repetitive or complex logic into separate helper functions or utilities.
- Simplify the component's interface by removing unnecessary props or state variables.
- Ensure that the component follows best practices for naming, formatting, and commenting.

By breaking down the Task Card component in this way, we can make it easier for developers to work with the code, and reduce the risk of bugs and errors creeping in over time.

TanishqSingla:
If the issue is resolved we can close this.

PR #627, opened 2023-06-11
refactor: Extract Assignee Button

TanishqSingla:
### Issue:
related to #526 

### Description:
Extract Assignee Button to separate component

### Dev Tested:
- [x] Yes

### DIFF
--- a/src/components/tasks/Card/AssigneeButton/AssigneeButton.tsx
+++ b/src/components/tasks/Card/AssigneeButton/AssigneeButton.tsx
@@ -1,0 +1,23 @@
+import { FC, HTMLProps } from 'react';
+
+import classNames from '@/components/tasks/Card/AssigneeButton/AssigneeButton.module.scss';
+interface Props extends HTMLProps<HTMLButtonElement> {
+    handleClick: () => Promise<void>;
+    assignee?: string;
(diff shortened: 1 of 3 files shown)


src/components/tasks/Card/AssigneeButton/AssigneeButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const AssigneeButton: FC<Props> = ({ handleClick, assignee, isDisabled }) => {
+    return (
+        <button
+            className={classNames.card__top__button}
+            type="button"
+            disabled={isDisabled}
+            onClick={handleClick}
+        >
+            {`Assign to ${assignee}`}
```
kotesh-arya, line 18: As you have mentioned  ` assignee?: string` where the assignee prop will be optional, in that case, what are we going to show as the Button text?  (`Assign to ${ ??? }`)

TanishqSingla: Its just following the current implementation, if the field is undefined, its gonna print `Assign to undefined` which is inline with the current implementation

sahsisunny (commented):
Please add the test coverage

PR #631, opened 2023-06-13
added footer to task details page

### LINKED ISSUE #614: Add Footer in the Task Details Page
Currently, the task details page lacks the footer as shown below:
image

bksh05:
### Issue:
There was no footer on the task details page.

### Description:
The task details page was missing the footer because the proper Layout was not being used.
Moved the Header code to a separate component.
Fixes #614 

N.A

### Dev Tested:
- [x] Yes

Before
image

image

After
image

image

https://github.com/Real-Dev-Squad/website-status/assets/27444637/c39cdc39-e17d-4e6d-aec3-90abd1e6badc

### DIFF
--- a/__tests__/Unit/Components/Header/Header.test.tsx
+++ b/__tests__/Unit/Components/Header/Header.test.tsx
@@ -1,0 +1,224 @@
+import { Header } from '@/components/Header';
+import { render, screen } from '@testing-library/react';
+
+/* eslint-disable */
+const useRouter = jest.spyOn(require('next/router'), 'useRouter');
+
+describe('Header without dev mode', () => {
+    it('should have Tasks category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Tasks');
+        expect(element).toBeInTheDocument();
+        expect(element?.classList.contains('active')).toBeTruthy();
+    });
+
+    it('should have Issues category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/issues',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Issues');
+        expect(element).toBeInTheDocument();
+        expect(element?.classList.contains('active')).toBeTruthy();
+    });
+
+    it('should have Mine category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/mine',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Mine');
+        expect(element).toBeInTheDocument();
+
+        expect(element?.classList.contains('active')).toBeTruthy();
+    });
+
+    it('should have Open PRs category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/openPRs',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Open PRs');
+        expect(element).toBeInTheDocument();
+
+        expect(element?.classList.contains('active')).toBeTruthy();
+    });
+
+    it('should have Stale PRs category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/stale-pr',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Stale PRs');
+        expect(element).toBeInTheDocument();
+
+        expect(element?.classList.contains('active')).toBeTruthy();
+    });
+
+    it('should have Idle Users category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/idle-users',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Idle Users');
+        expect(element).toBeInTheDocument();
+
+        expect(element?.classList.contains('active')).toBeTruthy();
+    });
+
+    it('should have Standup category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/standup',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Standup');
+        expect(element).not.toBeInTheDocument();
+    });
+
+    it('should have Availability Panel category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/availability-panel',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Availability Panel');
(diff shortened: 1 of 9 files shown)


src/components/Layout/index.tsx
```diff
@@ -9,6 +9,7 @@ import { Loader } from '../tasks/card/Loader';
 
 interface Props {
     children?: ReactNode;
+    hideHeader?: boolean; // Hides header when set to true
```
yesyash, line 12: - Comment is not needed here as the variable is descriptive

src/components/Layout/index.tsx
```diff
@@ -9,6 +9,7 @@ import { Loader } from '../tasks/card/Loader';
...
 }
 
 const navBarContent = (title: string, refUrl: string, isActive = false) => {
```
yesyash, line 15: - Please Change the first letter to a capital letter this is generally followed when building components in react.

src/components/Layout/index.tsx
```diff
@@ -36,54 +37,56 @@ const Layout: FC<Props> = ({ children }) => {
...
+                        )}
+                        {dev && (
+                            <>
+                                |
+                                {navBarContent(
+                                    'Availability Panel',
+                                    '/availability-panel'
+                                )}
+                            </>
+                        )}
```
yesyash, line 87: - When using a component we don't call it as a function, it should rather be used as `<NavBarContent />` this helps react keep track of your component. 

- This flow is very hard to maintain and read what if we rather create an array containing all navigation links and the map over them to render the footer?
Along with this, we can have a function that tells us if we have to render this component or not given a particular set of conditions, this would help us conditionally render the link the same way you're doing. 
What do you think?

yesyash (changes requested):

src/components/Layout/index.tsx
bksh05: That's correct. It can be written in a more readable way. 
But this was something already present and hence I did not fix it as part of this ticket. 
Do you want to fix it here? or create a new issue perhaps?

yesyash: - Even if we don't change the way the list is being rendered I would advise that we at least fix how the `navBarComponent` is being used here. And the rending of list can be taken up in a different PR.

kotesh-arya (commented):
Instead of using the **navBarContent** as a function expression we can replace it as a Functional component where ever it is used 👍🏻

src/components/Layout/index.tsx
bksh05: Done.

src/components/Layout/index.tsx
bksh05: I have extracted navBarComponent as a Header component.

yesyash (dismissed):

Pratiyushkumar (commented):
Request - please add a short video where tabs (task, issue, mine) are shown to be working fine just for reference of reviewers and future reference.

__tests__/Unit/Components/Header/Header.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+    it('should have Issues category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/issues',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Issues');
```
RitikJaiswal75, line 29: issues task is available by default in the header component I guess, so why are we testing it after changing the route, this test of queryByText of issues will pass even if it is written in line 16 I guess

__tests__/Unit/Components/Header/Header.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+    it('should have Mine category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/mine',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Mine');
+        expect(element).toBeInTheDocument();
```
RitikJaiswal75, line 43: same as above

__tests__/Unit/Components/Header/Header.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+    it('should have Open PRs category', () => {
+        useRouter.mockImplementation(() => {
+            return {
+                pathname: '/openPRs',
+                query: {},
+            };
+        });
+        render(<Header />);
+        const element = screen.queryByText('Open PRs');
+        expect(element).toBeInTheDocument();
```
RitikJaiswal75, line 57: same as above

src/components/Layout/index.tsx
```diff
@@ -1,73 +1,25 @@
...
-                                                    element.pathName
-                                            )}
-                                        </Fragment>
-                                    </>
-                                );
-                            })}
-                        </>
-                    )}
-                </div>
+                {!hideHeader ? <Header /> : null}
```
RitikJaiswal75, line 22: what do we do with this hideHeader

__tests__/Unit/Components/Header/Header.test.tsx
bksh05: These tests are to make sure that everything appears as it should be in both normal and dev mode.
Let's say some developer makes changes and by mistake Issues tab is not coming up anymore. Then it will fail and the developer will understand there is some mistake.

__tests__/Unit/Components/Header/Header.test.tsx
bksh05: Same as above

__tests__/Unit/Components/Header/Header.test.tsx
bksh05: Same as above

src/components/Layout/index.tsx
bksh05: The layout component by default shows the header. 
I am using the same Layout component to show the footer on the task details page. But on that page Header is not required.
And hence we have a prop to hide the header if we want to. 
This way we reuse the Layout component in the task details page as well.

src/interfaces/taskDetails.type.ts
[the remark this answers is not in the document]
bksh05: The types that I have created are in separate files as they should be according to the existing structure.
But there exist some interfaces/types which are not in separate files because some other developers chose to keep them there. 
Since it's not related to this ticket I am not touching those. 

If you want to move to separate files, then can we raise a separate issue for that?

__tests__/Unit/Components/Header/Header.test.tsx
bksh05: fixed.

yesyash (approved):

kotesh-arya (approved):

shubhamsigdar1 (approved):

RitikJaiswal75 (approved):

PR #632, opened 2023-06-13
refactored code to trigger less re-rendering

### LINKED ISSUE #605: Fix the code to render `IssueList` in `/issues` route.
### Description
The way we render Issues is not the best, it's hard for React to keep track of it and if the value in the `search field` updates all the components in `issueList` get re-rendered. 

A better way to implement this would have been to write the if-else logic where we place the variable to render this list. ([link to the line](https://github.com/Real-Dev-Squad/website-status/blob/develop/src/pages/issues/index.tsx#L117)). 

image

bksh05:
### Issue:
Fix the code to render IssueList in /issues route.

### Description:
On every input change on the search field, the issueList was being re-rendered. This was because the IssueList components was controlling the state of the SearchField component as well. 
I fixed this issue by refactoring the code and moving the search field-related text to the SearchField component.

Fixes #605 

### Dev Tested:
- [ ] Yes

Before the changes, the profiler shows it was re-rendered every time I typed something in the text box.
image

After the change, the profiler shows no re-rendering of the IssueList while changing the search field text

image

### DIFF
--- a/src/pages/issues/index.tsx
+++ b/src/pages/issues/index.tsx
@@ -52,7 +52,7 @@ const Issues: FC = () => {
     const [isLoading, setIsLoading] = useState<boolean>(false);
     const [error, setError] = useState<null | any>(null);
 
-    const fetchIssues = async (searchText: string) => {
+    const fetchIssues = async (searchText = '') => {
         try {
             setIsLoading(true);
             const res = await fetch(`${ISSUES_URL}?q=${searchText}`);
@@ -83,7 +83,7 @@ const Issues: FC = () => {
     };
 
     useEffect(() => {
-        fetchIssues('');
+        fetchIssues();
     }, []);
 
     let renderElement = <p>Loading...</p>;
--- a/src/pages/issues/index.tsx
+++ b/src/pages/issues/index.tsx
@@ -12,24 +12,22 @@ import { IssueItem } from '@/interfaces/issueItem.type';
 import { PullRequestAndIssueItem } from '@/interfaces/pullRequestIssueItem';
 
 type SearchFieldProps = {
-    searchText: string;
-    onSearchTextChanged: (event: ChangeEvent<HTMLInputElement>) => void;
-    onSearchTextSubmitted: () => void;
+    onSearchTextSubmitted: (searchString: string) => void;
     loading: boolean;
 };
 
-const SearchField = ({
-    searchText,
-    onSearchTextChanged,
-    onSearchTextSubmitted,
-    loading,
-}: SearchFieldProps) => {
+const SearchField = ({ onSearchTextSubmitted, loading }: SearchFieldProps) => {
+    const [searchText, setSearchText] = useState<string>('');
+    const onSearchTextChanged = (e: ChangeEvent<HTMLInputElement>) => {
+        setSearchText(e.target.value);
+    };
+
     return (
         <form
             className={classNames.searchFieldContainer}
             onSubmit={(e) => {
                 e.preventDefault();
-                onSearchTextSubmitted();
+                onSearchTextSubmitted(searchText);
             }}
         >
             <input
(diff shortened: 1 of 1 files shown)


src/pages/issues/index.tsx
```diff
@@ -12,24 +12,22 @@ import { IssueItem } from '@/interfaces/issueItem.type';
...
-    searchText,
-    onSearchTextChanged,
-    onSearchTextSubmitted,
-    loading,
-}: SearchFieldProps) => {
+const SearchField = ({ onSearchTextSubmitted, loading }: SearchFieldProps) => {
+    const [searchText, setSearchText] = useState<string>('');
+    const onSearchTextChanged = (e: ChangeEvent<HTMLInputElement>) => {
+        setSearchText(e.target.value);
+    };
```
yesyash, line 23: - There is another improvement we can make here because the search action is only performed by clicking the `submit` button.
    - Instead of storing the search value in the state which currently has no benefit, the search query can be taken from the form using `FormData` and passed on to the `onSearchTextSubmitted` function.

bksh05: The button should be disabled or not is checked using the searchText. For this, we need the state which updates only when the onChange event is fired. 
IMO, this part is already optimized, keeping user experience in mind.

Let me know if you think otherwise. I will be happy discuss on this if you feel we should optimise it further.

yesyash (dismissed):

shubhamsigdar1 (approved):
LGTM

src/pages/issues/index.tsx
```diff
@@ -86,12 +83,9 @@ const Issues: FC = () => {
     };
 
     useEffect(() => {
-        fetchIssues();
+        fetchIssues('');
     }, []);
```
Pratiyushkumar, line 87: why are we passing empty string over here

Pratiyushkumar: any specific reason?

Pratiyushkumar: don't you think we could have passed default argument as empty string instead of passing empty string as a parameter.

Pratiyushkumar (approved):
I am approving it but please take care of my comment as well please

src/pages/issues/index.tsx
bksh05: Yeah. It's a small change I will accommodate in some other PR. 
Thank you for your time

bksh05: Considering the fact that this change doesn't seem to enhance functionality or readability, I don't believe it is necessary to include it. Moreover, going through the approval process again for this change would considerably delay the agreed-upon timelines.

However, I want to assure you that I will address this issue in my next pull request on the status repository.

I hope this rationale aligns with your understanding and that you agree with my thought process.

Pratiyushkumar: Also, I guess you would need to take re-approvals because today some changes were merged in the develop branch in the Layout file so request you to please take a fresh pull and then resolve the merge conflicts and make necessary changes and the changes that I have suggested

Pratiyushkumar (dismissed):
LGTM

RitikJaiswal75 (changes requested):
Please put in the test coverage report, in case if there is none, please write them

bksh05:
> Please put in the test coverage report, in case if there is none, please write them

image

@RitikJaiswal75 The test is above 85% of the component that I worked with. 
My changes do not add any functionality. It's just some refactoring to reduce re-rendering. Hence I did not write a test for that.

shubhamsigdar1 (approved):
Great work @bksh05

RitikJaiswal75:
Are these multiple re-renders visible visually if yes please add a video of that, can you please add a video of changes after your code change

bksh05:
> Are these multiple re-renders visible visually if yes please add a video of that, can you please add a video of changes after your code change

These are not visible visually on the screen. you can see them only when you run performance tools. 
After fixing I have pasted a screenshot which shows it is not re-rendering anymore.

yesyash (approved):

RitikJaiswal75 (approved):

PR #642, opened 2023-06-17
Tests for User-Suggestions

fakhruddinkw:
This tests are for #575 

**TEST COVERAGE**

test-coverage

### DIFF
--- a/__tests__/Unit/Components/Tasks/SuggestionBox.test.tsx
+++ b/__tests__/Unit/Components/Tasks/SuggestionBox.test.tsx
@@ -1,0 +1,65 @@
+import SuggestionBox from '@/components/tasks/SuggestionBox/SuggestionBox';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { DUMMY_PROFILE as placeholderImageURL } from '@/constants/display-sections';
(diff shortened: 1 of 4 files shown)


iamitprakash (approved):

__tests__/Unit/Components/Tasks/SuggestionBox.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+        const listImages = screen.getAllByTestId('image');
+
+        expect(list).toBeInTheDocument();
+        expect(listItems).toHaveLength(2);
+        expect(listImages).toHaveLength(2);
+
+        fireEvent.click(listItems[0]);
+        expect(onClickName).toHaveBeenCalled();
+    });
+});
```
bharati-21, line 65: Can you also please add a test for a scenario where you'd have an error?

fakhruddinkw: @bharati-21 do you mean in the scenario where my onClickname fails?

bharati-21: No as in the fetch call for the users fail

PR #655, opened 2023-06-23
Fix : inconsistency in the width of task-tabs and task-cards

j24m:
### Description: 
Fixed the inconsistency in the width of task-tabs and task-cards.

### Dev Tested:
- [x] Yes

https://github.com/Real-Dev-Squad/website-status/assets/53934353/e3ac9266-c47e-4e77-9056-3ea3254e00a6

### DIFF
--- a/src/components/Header/Header.module.scss
+++ b/src/components/Header/Header.module.scss
@@ -1,15 +1,19 @@
 .header {
+    margin-top: 0.5rem;
     display: flex;
     justify-content: center;
-    align-items: center;
     flex-wrap: wrap;
-    padding-top: 0.625rem;
+    padding: 1rem;
+    align-items: center;
+    gap: 1rem;
+    border-bottom: 1px solid var(--toastify-text-color-light);
 }
 
 .link {
-    padding: 0.625rem;
+    padding: 0.5rem;
+    width: 6rem;
     text-decoration: none;
-    font-weight: bold;
+    font-size: 14px;
     color: #041484;
     cursor: pointer;
     background: none;
--- a/src/components/Header/Header.module.scss
+++ b/src/components/Header/Header.module.scss
@@ -1,18 +1,16 @@
 .header {
-    margin-top: 0.5rem;
+    padding-top: 0.6rem;
     display: flex;
     justify-content: center;
     flex-wrap: wrap;
-    padding: 1rem;
     align-items: center;
-    gap: 1rem;
     border-bottom: 1px solid var(--toastify-text-color-light);
 }
 
 .link {
-    padding: 0.5rem;
+    padding: 0.6rem;
     text-decoration: none;
-    font-size: 14px;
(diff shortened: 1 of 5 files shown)


src/components/Header/Header.module.scss
```diff
@@ -1,15 +1,19 @@
...
+    border-bottom: 1px solid var(--toastify-text-color-light);
 }
 
 .link {
-    padding: 0.625rem;
+    padding: 0.5rem;
+    width: 6rem;
     text-decoration: none;
-    font-weight: bold;
+    font-size: 14px;
```
sahsisunny, line 16: Can you use REM instead of PX if the value is less than 8px?

src/components/tasks/card/card.module.scss
```diff
@@ -35,7 +36,7 @@ $seeMoreTasksShadow: #595959cc;
     justify-content: space-between;
     border: 1px solid #d9d9d9;
     position: relative;
-    max-width: 720px;
+    width: 1050px;
```
sahsisunny, line 39: Width too wide can you decrease the width?

harshith-venkatesh: How are we handling responsiveness here

j24m: This issue was raised to fix the inconsistent width but okay I will handle the responsiveness as well.

j24m: > Width too wide can you decrease the width?
I changed it on purpose to match it with the task-tab-container width.

j24m: > How are we handling responsiveness here

This issue was raised to fix the inconsistent width but okay I will handle the responsiveness as well.

j24m: > Width too wide can you decrease the width?

Design wise it matches the width of task-tabs-container and hence the change.

src/components/Header/Header.module.scss
j24m: 1rem = 16px, So if I'm not wrong we can use rem when greater than or equal to 16px. But in this case 14px = 0.875 rem.

kotesh-arya (approved):

src/components/Header/Header.module.scss
harshith-venkatesh: Could you share resources for the above statement please

kotesh-arya (dismissed):

kotesh-arya (approved):
Looks good to merge 👍🏻

harshith-venkatesh (approved):
Good work in resolving the issue, I believe all the screens are thoroughly checked and tested.
Approving from my end

j24m:
> Good work in resolving the issue, I believe all the screens are thoroughly checked and tested. Approving from my end

Thankyou :)

Issue #676, opened 2023-07-02
PullRequestList - @RitikJaiswal75


RitikJaiswal75:
Create a RTK query for fetching prs

PR #677, opened 2023-07-02
add rtk query for open prs api

RitikJaiswal75:
### Issue:
Migration of Pull request API to RTK Query #676 

### Description:
create a pull request API using rtk query and write its tests

### Dev Tested:
- [x] Yes

Test Coverage
image

### DIFF
--- a/__mocks__/handlers/pull-requests.handler.ts
+++ b/__mocks__/handlers/pull-requests.handler.ts
@@ -2,13 +2,10 @@ import { rest } from 'msw';
 import { mockGetOpenPrs, mockGetStalePrs } from '../db/prs';
 const URL = process.env.NEXT_PUBLIC_BASE_URL;
 
-export const openPrsHandler = [
+export const prsHandler = [
     rest.get(`${URL}/pullrequests/open?page=1&size=3`, (_, res, ctx) => {
         return res(ctx.status(200), ctx.json(mockGetOpenPrs));
     }),
-];
-
-export const stalePrsHandler = [
     rest.get(`${URL}/pullrequests/stale?page=1&size=3`, (_, res, ctx) => {
         return res(ctx.status(200), ctx.json(mockGetStalePrs));
     }),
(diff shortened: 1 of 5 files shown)


__mocks__/handlers/pull-requests.handler.ts
```diff
@@ -0,0 +1,15 @@
...
+    rest.get(`${URL}/pullrequests/open?page=1&size=3`, (_, res, ctx) => {
+        return res(ctx.status(200), ctx.json(mockGetOpenPrs));
+    }),
+];
+
+export const stalePrsHandler = [
+    rest.get(`${URL}/pullrequests/stale?page=1&size=3`, (_, res, ctx) => {
+        return res(ctx.status(200), ctx.json(mockGetStalePrs));
+    }),
+];
```
bhtibrewal, line 15: we don't need to make 2 separate handlers for a single API
`getPrsHandler`  and both will go in it

bhtibrewal (changes requested):
left a comment

prakashchoudhary07 (approved):

__mocks__/handlers/pull-requests.handler.ts
RitikJaiswal75: Handlers are just used for tests in mocks so I think they will be helpful for people who want to understand what scenarios will this query be used for

RitikJaiswal75: Since the status site has 2 pages based on PRs 1 for stale and one for open keeping 2 handlers in tests bifurcates and explicitly showing both the use case can be implemented.

RitikJaiswal75: thanks for explaining it. Please re-review.

yesyash (approved):

prakashchoudhary07 (approved):

bhtibrewal (approved):
LGTM

Issue #694, opened 2023-07-06
RTK Query: migrate the fetch call of user-suggestion component to RTK Query


fakhruddinkw:
As we are using RTK Query for all of our API calls, need to remove fetchUsers from user-suggestion box and use RTK Query for getting user's names.

This is regarding PR : #575 and #634

PR #703, opened 2023-07-10
replaced fetchUsers with useGetAllUsersByUsernameQuery

fakhruddinkw:
### Issue:

This PR is regarding: #694 

- [x] Replaced fetchUsers with useGetAllUsersByUsernameQuery hook
- [x] Need to implement debounce 
- [x] Tests

**Test Coverage**
useDebounce
Suggestions


bhtibrewal (changes requested):
please write tests for the changes and add the stats in decsription

src/components/tasks/SuggestionBox/Suggestions.tsx
```diff
@@ -0,0 +1,75 @@
...
+        ref
+    ) => {
+        const { data, isLoading } = useGetAllUsersByUsernameQuery(
+            {
+                searchString: searchTerm,
+            },
+            { skip: searchTerm ? false : true }
+        );
+
+        const usersData = data?.users;
```
sumitd94, line 30:
```suggestion
        const { users } = data;
```

src/components/tasks/SuggestionBox/Suggestions.tsx
```diff
@@ -0,0 +1,75 @@
...
+        const suggestedUsers: GithubInfo[] = [];
+
+        usersData?.map((data: userDataType) => {
```
sumitd94, line 33:
```suggestion
        users?.map((user: userDataType) => {
```

src/components/tasks/SuggestionBox/Suggestions.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+        return (
+            <div className={classNames.suggestionDiv}>
+                <input
+                    ref={ref}
+                    value={assigneeName}
+                    className={classNames.cardStrongFont}
+                    onKeyDown={(e) => {
+                        handleChange(e, 'assignee');
+                    }}
```
sumitd94, line 50:
```suggestion
                    onKeyDown={(e) => handleChange(e, 'assignee')}
```

src/components/tasks/SuggestionBox/Suggestions.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    onChange={(e) => {
+                        handleAssignment(e);
+                    }}
```
sumitd94, line 53: same as above

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,19 @@
...
+
+        return () => {
+            clearTimeout(handler);
+        };
+    }, [value, delay]);
+
+    return debouncedValue;
+}
+
+export default useDebounce;
```
sumitd94, line 19: tests for this?

src/components/tasks/SuggestionBox/Suggestions.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    )
+                )}
+            </div>
+        );
+    }
+);
+
+Suggestions.displayName = 'Suggestions';
+
+export default Suggestions;
```
sumitd94, line 75: tests missing

src/components/tasks/SuggestionBox/Suggestions.tsx
fakhruddinkw: I have to do optional chaining, because otherwise it throws error that the userResponse doesnt exist

__mocks__/handlers/users.handler.js
```diff
@@ -12,46 +12,49 @@ const usersHandler = [
...
-            users: filteredUsers,
-            links: {
-                next: nextId ? `/users?size=${size}&next=${nextId}&search=${searchParamToMatch}` : '',
-                prev: prevId ? `/users?size=${size}&prev=${prevId}&search=${searchParamToMatch}` : '',   
-            }
+                message: 'Users returned successfully!',
+                users: filteredUsers,
+                links: {
+                    next: nextId
+                        ? `/users?size=${size}&next=${nextId}&search=${searchParamToMatch}`
```
bharati-21, line 49: Nit: can create a const variable and then only append next or prev instead of typing the entire URL twice.

__tests__/Unit/Components/Tasks/Suggestions.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+        );
+        expect(screen.getByRole('button')).toBeInTheDocument();
+
+        const inputEl = screen.getByRole('button');
+        const event = { target: { value: 'ab' } };
+        fireEvent.change(inputEl, event as React.ChangeEvent<HTMLInputElement>);
+
+        expect(handleAssignment).toHaveBeenCalledTimes(1);
+    });
+});
```
bharati-21, line 88: Can you please test an error scenario when the call fails to load suggestions

__tests__/Unit/hooks/useDebounce.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    });
+
+    test('should return the updated value after delay', async () => {
+        renderHook(() => useDebounce('hello world', 500));
+
+        // Second render with "world"
+
+        await waitFor(() => {
+            const { result } = renderHook(() => useDebounce('pratiyush', 500));
+            jest.advanceTimersByTime(300); // Move the timers forward by another 200ms, total 500ms
```
bharati-21, line 28: Move the timers forward by another 200ms, total 500ms, but you passed 300 here `jest.advanceTimersByTime(300)` - can you please explain this?

__tests__/Unit/hooks/useDebounce.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+            expect(result.current).toBe('pratiyush');
+        });
+
+        await waitFor(() => {
+            const { result } = renderHook(() => useDebounce('world', 500));
+            jest.advanceTimersByTime(200); // Move the timers forward by another 200ms, total 500ms
```
bharati-21, line 34: And here you passed 200

__tests__/Unit/hooks/usersApi.test.tsx
```diff
@@ -124,3 +125,30 @@ describe('useGetUsersByUsernameQuery', () => {
...
+        const users = nextResponse.data?.users;
+        expect(nextResponse.data).not.toBeUndefined();
+        expect(users).toHaveLength(8);
+        expect(nextResponse.isLoading).toBe(false);
+        expect(nextResponse.isSuccess).toBe(true);
+        users?.forEach((user) => {
+            expect(user.username).toMatch(/^mu/);
+        });
+    });
+});
```
bharati-21, line 154: Please write test case to match error scenario and when no user match is found.

src/components/tasks/SuggestionBox/Suggestions.tsx
```diff
@@ -0,0 +1,72 @@
...
+    (
+        {
+            assigneeName,
+            searchTerm,
+            showSuggestion,
+            handleAssignment,
+            handleChange,
+            handleClick,
+        },
+        ref
```
bharati-21, line 21: Why is ref being passed as props ?

src/components/tasks/card/index.tsx
```diff
@@ -529,37 +487,15 @@ const Card: FC<CardProps> = ({
...
-                                )}
-                            </div>
+                            <Suggestions
+                                assigneeName={assigneeName}
+                                searchTerm={debounceSearchTerm}
+                                showSuggestion={showSuggestion}
+                                handleAssignment={handleAssignment}
+                                handleClick={handleClick}
+                                handleChange={handleChange}
+                                ref={inputRef}
```
bharati-21, line 497: Are these handlers  & state used anywhere within card component? If not, please move it to the Suggestions component to avoid adding unwanted handlers and state. 

For e.g. showSuggestion, I don't see the use of this state anywhere within this card component. 
Did you check if this change creates any unused state and set state and aother handlers if you don't pass them as props. If yes, they aren't required here. Please ensure this and only then keep them here.

bharati-21: I don't see assigneeName, setAssigneeName being used here. Even the click handlers, debounced search term - please move them to the specific component.

bharati-21 (changes requested):
Left some comments, kindly look into those.

__tests__/Unit/hooks/useDebounce.test.tsx
fakhruddinkw: sorry that was old comments, which i forgot to remove @bharati-21

__tests__/Unit/hooks/useDebounce.test.tsx
fakhruddinkw: ignore these comments, sorry @bharati-21

src/components/tasks/SuggestionBox/Suggestions.tsx
fakhruddinkw: because i have a ref on input element, but i am accessing it in the parent element. @bharati-21

src/components/tasks/card/index.tsx
fakhruddinkw: @bharati-21 showSuggestion is being passed as prop to Suggestions, but it is being toggled in handleAssignment and handleClick thats why they reside in the index.tsx instead of in Suggestions. 
Also setAssigneeName is also being used in handleAssignment in index.tsx

PR #709, opened 2023-07-13
Revert "fixed editing of title field"

fakhruddinkw:
Reverts Real-Dev-Squad/website-status#706

replacing assigneeName with event.target.innerHTML is sending empty string which throws an error of 'assignee' cannot be empty.


Pratiyushkumar (approved):
LGTM

sahsisunny (approved):

PR #723, opened 2023-07-16
FEATURE: Add intersection observer for fetching task

sahsisunny:
### Description:
As part of this implementation, we will use the Intersection Observer API to automatically fetch the next task when the user reaches the end of the task list container. This will eliminate the need for the "Load More" button, providing a smoother user experience.

### Dev Tested:
- [x] Yes

### Test Coverage:
image
image

https://github.com/Real-Dev-Squad/website-status/assets/70854507/fc0ae344-ad0b-4e61-bd4c-21033bcd474e

### DIFF
--- a/src/components/tasks/TasksContent.tsx
+++ b/src/components/tasks/TasksContent.tsx
@@ -1,3 +1,4 @@
+import { ElementRef } from 'react';
 import classNames from '@/styles/tasks.module.scss';
 import { useGetAllTasksQuery } from '@/app/services/tasksApi';
 import task, { TABS, Tab, TabTasksData } from '@/interfaces/task.type';
@@ -66,6 +67,8 @@ export const TasksContent = ({ dev }: { dev: boolean }) => {
         MERGED: [],
         COMPLETED: [],
     });
+    const loadingRef = useRef<ElementRef<'div'>>(null);
+    const bottomBoundaryRef = useRef<ElementRef<'div'>>(null);
 
     const {
         data: tasksData = { tasks: [], next: '' },
@@ -109,9 +112,6 @@ export const TasksContent = ({ dev }: { dev: boolean }) => {
             setLoadedTasks(newTasks);
         }
     }, [tasksData.tasks]);
-
-    const loadingRef = useRef<HTMLDivElement | null>(null);
-    const bottomBoundaryRef = useRef<HTMLDivElement | null>(null);
 
     useIntersection(loadingRef, bottomBoundaryRef, fetchMoreTasks);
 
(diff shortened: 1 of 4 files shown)


src/hooks/useIntersection.ts
```diff
@@ -0,0 +1,56 @@
+import { useEffect, useState } from 'react';
+const useIntersection = (
+    loadingRef: React.MutableRefObject<HTMLDivElement | null>,
+    bottomBoundaryRef: React.MutableRefObject<HTMLDivElement | null>,
+    onLoadMore: () => void
+) => {
+    const [isIntersecting, setIsIntersecting] = useState(false);
+
+    useEffect(() => {
+        const options = {
```
apurvkhare, line 10: do we need to define the `options` constant in a `useEffect`? It could very well be outside useIntersection.

src/hooks/useIntersection.ts
```diff
@@ -0,0 +1,56 @@
...
+        };
+    }, [loadingRef]);
+
+    useEffect(() => {
+        if (isIntersecting) {
+            onLoadMore();
+        }
+    }, [isIntersecting, onLoadMore]);
+
+    useEffect(() => {
```
apurvkhare, line 37: Why is the scroll event also required for this hook? IntersectionObsever alone should suffice.

src/components/tasks/TasksContent.tsx
```diff
@@ -109,6 +110,11 @@ export const TasksContent = ({ dev }: { dev: boolean }) => {
         }
     }, [tasksData.tasks]);
 
+    const loadingRef = useRef<HTMLDivElement | null>(null);
```
apurvkhare, line 113: to avoid `| null` the `useRef` can be defined like below
```
import {ElementRef} from 'react'

const loadingRef = useRef<ElementRef<"div">>(nulll);
const bottomBoundaryRef = useRef<ElementRef<"div">>(null);
```

FYI, your type definition is also fine, this is just a suggestion.

src/hooks/useIntersection.ts
sahsisunny: IntersectionObserver has some limitations, and using both approaches ensures better handling of the load more functionality.

src/components/tasks/TasksContent.tsx
sahsisunny: Done

src/hooks/useIntersection.ts
sahsisunny: Done

Pratiyushkumar (approved):
LGTM

yesyash (approved):

Pratiyushkumar (approved):

sahsisunny:
> useIntersection hook is using a intersection observer and a scroll event where the scroll event is not required as we're using the intersection observer already.

Not able to find an alternative solution and this approach does not have any issue

vinit717 (approved):
Just look for a good approach if you can find it later for now this seems good to me as no other is there is now

PR #728, opened 2023-07-17
changed design of dropdown

fakhruddinkw:
This is regarding: https://github.com/Real-Dev-Squad/todo-action-items/issues/151

**OLD DESIGN**
old-drop-down-status

**NEW DESIGN**
status-new

**RESPONSIVENESS**

https://github.com/Real-Dev-Squad/website-status/assets/85249044/2f9eae72-aabc-416f-832b-0fad933f8658

**COVERAGE**
dropdown

PR #616, opened 2023-06-06
REFACTOR: Move the Update-progress button behind feature flag

kotesh-arya:
### Issue:
 Set the UPDATE PROGRESS feature behind a feature flag  https://github.com/Real-Dev-Squad/website-status/issues/612

### Description: 
Moving the feature **Update Progres** behind a feature flag as this feature is not completed yet so moving this button behind the feature flag and hiding it until this feature is completely built.

### Dev Tested:
- [x] Yes

### Test Coverage Stats:
image
image

### Images/video of the change:
Without Feature-Flag: 
image
With Feature-Flag: 
image

PR #734, opened 2023-07-18
Prevent Hello with empty profile rendering in navbar when user already loggedin

kotesh-arya:
### Issue:
https://github.com/Real-Dev-Squad/website-status/issues/746
### Description:
`Login with Github` button showing have been handled in the Navbar while refreshing the Pages, when user already Loggedin
### Dev Tested:
- [x] Yes

**Before:**

https://github.com/Real-Dev-Squad/website-status/assets/69259490/24548937-15bb-466a-86ff-aaf1cc0395fc

**After:**

https://github.com/Real-Dev-Squad/website-status/assets/69259490/1a95ab8b-fc42-4bca-8226-039bb9d3f4cf

### Test Coverage Stats:

image

### DIFF
--- a/__tests__/Unit/Components/Navbar/Navbar.test.tsx
+++ b/__tests__/Unit/Components/Navbar/Navbar.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { screen, render } from '@testing-library/react';
+import { screen, render, waitFor } from '@testing-library/react';
 import NavBar from '../../../../src/components/navBar';
 import * as authHooks from '@/hooks/useAuthenticated';
 import { renderWithProviders } from '@/test-utils/renderWithProvider';
@@ -90,5 +90,13 @@ describe('Navbar', () => {
         await screen.findAllByTestId('navbar');
         const logo = getByTestId('logo');
         expect(logo).toBeInTheDocument();
+    });
+    test('Sign In With Github button should not render when user is already loggedin', async () => {
+        renderWithProviders(<NavBar />);
+        await waitFor(() => {
+            const loginTextElement = screen.queryByText('Sign In With Github');
+            expect(loginTextElement).not.toBeInTheDocument();
+            screen.debug();
+        });
     });
 });
--- a/__tests__/Unit/Components/Navbar/Navbar.test.tsx
+++ b/__tests__/Unit/Components/Navbar/Navbar.test.tsx
@@ -96,7 +96,6 @@ describe('Navbar', () => {
         await waitFor(() => {
             const loginTextElement = screen.queryByText('Sign In With Github');
             expect(loginTextElement).not.toBeInTheDocument();
-            screen.debug();
         });
     });
 });
--- a/src/components/navBar/index.tsx
+++ b/src/components/navBar/index.tsx
@@ -78,40 +78,44 @@ const NavBar = () => {
                 </ul>
             </div>
             <div className={styles.userProfile}>
-                {!isLoggedIn ? (
-                    <Link href={LOGIN_URL}>
-                        <button className={styles.signInLink}>
-                            Sign In With Github
+                {userData !== undefined ? (
+                    userData ? (
+                        <div
+                            className={styles.userGreet}
+                            onClick={() => setToggleDropdown(!toggleDropdown)}
+                        >
+                            <div className={styles.userGreetMsg}>
+                                Hello, {userData?.first_name}
+                            </div>
                             <Image
-                                className={styles.githubLogo}
-                                src={GITHUB_LOGO}
-                                height="20"
-                                width="20"
-                                alt="GitHub Icon"
+                                className={styles.userProfilePic}
+                                src={
+                                    userData?.picture?.url
+                                        ? `${userData?.picture?.url}`
+                                        : `${DEFAULT_AVATAR}`
+                                }
+                                alt="Profile Pic"
+                                width="32"
+                                height="32"
                             />
-                        </button>
-                    </Link>
-                ) : (
-                    <div
-                        className={styles.userGreet}
-                        onClick={() => setToggleDropdown(!toggleDropdown)}
-                    >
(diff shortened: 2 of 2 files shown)


src/components/taskDetails/index.tsx
```diff
@@ -238,13 +239,17 @@ const TaskDetails: FC<Props> = ({ taskID }) => {
...
+                                <button
+                                    onClick={() =>
+                                        router.push(
+                                            `/progress/${taskID}?dev=true`
+                                        )
+                                    }
+                                >
+                                    Update Progress
+                                </button>
+                            )}
```
vinit717, line 252: why this changes reflect in this pr

kotesh-arya: Actually, there's another PR which is yet to be merged, once it gets merged, these changes will no more exist 👍🏻 
Here's the PR: https://github.com/Real-Dev-Squad/website-status/pull/616

src/components/taskDetails/index.tsx
```diff
@@ -56,6 +56,8 @@ type Props = {
 
 const TaskDetails: FC<Props> = ({ taskID }) => {
     const router = useRouter();
+    const { query } = router;
+    const isDevModeEnabled = query.dev === 'true' ? true : false;
```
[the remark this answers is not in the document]
kotesh-arya: > If the possible values of `query.dev` are only 'true' and 'false' then you can choose to use the below code.
> 
> ```
> JSON.parse(query.dev);
> ```
> 
> This is just another way of converting 'true' and 'false' strings to respective booleans. It is up to you if you want to use this, the conditional operator code is understandable. If you use JSON.parse add a brief comment on what it does.

Thanks for the great suggestion @apurvkhare  🙌🏻 
But in this case I am targeting only for `query.dev` value being **true** when it has string 'true'  and in all the rest cases of `query.dev` will be false.

src/components/navBar/index.tsx
```diff
@@ -78,40 +78,44 @@ const NavBar = () => {
                 </ul>
             </div>
             <div className={styles.userProfile}>
-                {!isLoggedIn ? (
-                    <Link href={LOGIN_URL}>
-                        <button className={styles.signInLink}>
-                            Sign In With Github
+                {userData !== undefined ? (
+                    userData ? (
```
TanishqSingla, line 82: if userData happens to be `{}` it will return true, make sure you're handling that case

src/components/navBar/index.tsx
```diff
@@ -78,40 +78,44 @@ const NavBar = () => {
...
+                                    src={GITHUB_LOGO}
+                                    height="20"
+                                    width="20"
+                                    alt="GitHub Icon"
+                                />
+                            </button>
+                        </Link>
+                    )
+                ) : (
+                    ''
```
TanishqSingla, line 118: we can just rely on `&&` operator for conditional rendering, that way we won't have to print `''`

TanishqSingla (changes requested):

src/components/taskDetails/index.tsx
kotesh-arya: **P.S: ** Now those changes will no more reflect in this PR.

src/components/navBar/index.tsx
kotesh-arya: Updated with Short Circuit Operator - **&&** 👍🏻

__tests__/Unit/Components/Navbar/Navbar.test.tsx
```diff
@@ -91,4 +91,12 @@ describe('Navbar', () => {
         const logo = getByTestId('logo');
         expect(logo).toBeInTheDocument();
     });
+    test('Sign In With Github button should not render when user is already loggedin', async () => {
+        renderWithProviders(<NavBar />);
+        await waitFor(() => {
+            const loginTextElement = screen.queryByText('Sign In With Github');
+            expect(loginTextElement).not.toBeInTheDocument();
+            screen.debug();
```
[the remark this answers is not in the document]
kotesh-arya: Removed, Thanks for pointing 🤝

src/components/navBar/index.tsx
kotesh-arya: The userData will either be undefined or an Object with Data properties right ?
Can you please elaborate on this empty Object { } case, Thank you

bksh05 (approved):
LGTM

src/components/navBar/index.tsx
```diff
@@ -78,41 +78,42 @@ const NavBar = () => {
                 </ul>
             </div>
             <div className={styles.userProfile}>
-                {!isLoggedIn ? (
-                    <Link href={LOGIN_URL}>
-                        <button className={styles.signInLink}>
-                            Sign In With Github
+                {userData !== undefined &&
```
TanishqSingla, line 81: instead of checking equality with undefined, we can just use `userData` in this line, this will return a falsy value. If you want to be more specific about the boolean then you can use it with `!!userData` which will convert the Falsy value to `false`

TanishqSingla (changes requested):

src/components/navBar/index.tsx
kotesh-arya: Updated as per the suggestion 👍🏻

TanishqSingla (approved):

src/components/navBar/index.tsx
```diff
@@ -78,41 +78,42 @@ const NavBar = () => {
...
-                            }
-                            alt="Profile Pic"
-                            width="32"
-                            height="32"
-                        />
-                        {toggleDropdown && <Dropdown />}
-                    </div>
-                )}
+                    ) : (
+                        <Link href={LOGIN_URL}>
```
shubhamsigdar1, line 104: When it will run??

kotesh-arya:
Closing this request as this issue had been addressed in another PR https://github.com/Real-Dev-Squad/website-status/pull/728

Issue #497, opened 2023-04-14
Refactor Task Card Component


bhtibrewal:
This is still a massive chunk of JSX which is hard to follow. We should be breaking this down into meaningful individual UI components.

_Originally posted by @hjaintech in https://github.com/Real-Dev-Squad/website-status/pull/468#discussion_r1162360142_

TanishqSingla:
Is there anything remaining in this PR? If not can we close this?

PR #736, opened 2023-07-18
Fix edit button

bhtibrewal:
### Description: 
- when clicking on the edit button the complete page reloads and we loss focus from the task we wanted to edit.
- removed page reload caused due to query param 

### Dev Tested:
- [x] Yes

image

### Refactoring 
https://github.com/Real-Dev-Squad/website-status/issues/497


src/components/tasks/card/index.tsx
```diff
@@ -72,6 +72,12 @@ const Card: FC<CardProps> = ({
 
     const [keyLongPressed] = useKeyLongPressed();
 
+    const [isEditMode, setIsEditMode] = useState(false);
+    const onEditRoute = () => {
+        setIsEditMode(true);
+    };
```
yesyash, line 78: - Move this below the `useEffect` ? As we generally have this hierarchy in a react component where we: 
```js
-- declare your state / hooks --
-- variables dependent on state / value from hooks --
-- event / action handlers --
-- jsx mark / return --

**Review by yesyash — CHANGES_REQUESTED** [review]

File: src/components/tasks/card/index.tsx [hunk 2]
```diff
@@ -72,6 +72,12 @@ const Card: FC<CardProps> = ({
...
+    shouldEdit = shouldEdit && isUserAuthorized && isEditMode;
```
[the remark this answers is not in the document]
bhtibrewal: no, it's not always true, in other places where this card is being used it's set to false

src/components/tasks/card/index.tsx
```diff
@@ -242,6 +237,11 @@ const Card: FC<CardProps> = ({
             });
     };
 
+    const onEditRoute = () => {
+        setIsEditMode(true);
+    };
+    shouldEdit = shouldEdit && isUserAuthorized && isEditMode;
```
yesyash, line 243: - I think having this boolean variable near `isEditMode` would be better as it will be clear that the of `isEditMode` is used to make a decision.

src/components/tasks/TaskList/TaskList.tsx
```diff
@@ -62,7 +56,7 @@ export default function TaskList({ tasks, hasLimit = false }: TaskListProps) {
                 <Card
                     content={item}
                     key={item.id}
-                    shouldEdit={isEditable}
+                    shouldEdit
```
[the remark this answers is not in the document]
bhtibrewal: as I stated in the below comment, that the prop shouldEdit is set to false so at the other places it being used it's not editable,  so here just passing it is same as shouldEdit= true

src/components/tasks/card/index.tsx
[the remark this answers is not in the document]
yesyash: > no, it's not always true, in other places where this card is being used it's set to false

- in that case why are we directly modifying the `shouldEdit` prop ? Should this not be a new variable instead?

src/components/tasks/card/index.tsx
bhtibrewal: disn't understand your pount here, can you please explain

yesyash: - having the state and it's dependent variables together makes it easier to read and made future edits to the code.
keeping this in mind should we move the block below near `isEditMode` useState? 
```js
shouldEdit = shouldEdit && isUserAuthorized && isEditMode;
```

bharati-21: Also `shoulEdit` is props coming from the parent. Props should not be modified by the child components.

src/components/tasks/card/index.tsx
```diff
@@ -473,12 +473,12 @@ const Card: FC<CardProps> = ({
                             Estimated completion
                         </span>
                         <span className={classNames.completionDate}>
-                            {renderDate(fromNowEndsOn, shouldEdit)}
+                            {renderDate(fromNowEndsOn, isEditable)}
```
yesyash, line 476: - as we're returning JSX here will it be better to use this as a react component and not a function call ?

bhtibrewal: agreed with you on this but is a separate refactor task, doesn't come under the scope of this PR

yesyash: - viewing the file I am not able to see the `renderDate` being used it in once place that is here, can we do this in this pr as it does not add much to the scope?

bhtibrewal: No yash it's not just rename if I rename this there are other minor refactors also, and I don't want to address them in this PR, refactor is being addressed separately

yesyash: - got it, in that case can you please link the refactor ticket in the pr.

bhtibrewal: https://github.com/Real-Dev-Squad/website-status/issues/497

yesyash (approved):

vvaibhavdesai (approved):

PR #743, opened 2023-07-20
make userProgressQuery to getProgressQuery

RitikJaiswal75:
## Overview

This PR aims to refactor the existing `userProgressDetails` RTK query to make it more generic and reusable. The query will be renamed as `getProgressDetails` to support fetching progress details for both users and tasks.

## Changes Made

The following changes have been made to this pull request:

- Refactored the `userProgressDetails` RTK query to make it generic and renamed it to `getProgressDetails`.
- Updated all references to `userProgressDetails` to use the new `getProgressDetails` query.
- Added comprehensive unit tests to validate the correctness and versatility of the `getProgressDetails` query.
- Updated documentation and examples to reflect the changes and provide guidance on using the new query.

## Reasoning

The original `userProgressDetails` query was designed specifically for retrieving user progress details, which limited its applicability. By making it more generic, we can now use the `getProgressDetails` query to fetch progress details for both users and tasks, reducing code duplication and improving maintainability.

## Testing

We have added extensive unit tests to cover various scenarios for the `getProgressDetails` query. The tests validate the behavior of the query for both user progress and task progress cases, ensuring its correct functionality in any usage scenario.

## How to Test

To test the changes made in this pull request, follow these steps:

1. Checkout this branch locally.
2. Run the test suite using the following command:


src/app/services/progressesApi.ts
```diff
@@ -1,5 +1,10 @@
 import { api } from './api';
-import { userDetails } from '@/types/standup.type';
+import { progressDetails } from '@/types/standup.type';
+
+type queryParamsType = {
```
apurvkhare, line 4: If I understand correctly, in one request either taskId will be a param or userId, not both.
The current type does not enforce that.
Below type will enforce it
```
type queryParamsType = {
    userId: string;
} |  {
    taskId: string;
} 
```
If both params can be passed together then ignore this comment.

heyrandhir (approved):

src/app/services/progressesApi.ts
RitikJaiswal75: I am not sure how would I handle it in the place where I am using it below.

RitikJaiswal75: because with this i will have to make 2 RTK Queries making me write the repetitive code again

src/app/services/api.ts
```diff
@@ -24,7 +24,7 @@ export const api = createApi({
         'Levels',
         'Challenges',
         'Idle_Members',
-        'User_Standup',
```
sahsisunny, line 27: Why are you removed?

sahsisunny (approved):

src/app/services/progressesApi.ts
apurvkhare: I guess if you change your code like below it should work

```
query:  (param: queryParamsType) => {
return {
                 url: '/progresses',
                 params: param,
           };
}
```

Still if you this helps, else no issues with the current code.

src/app/services/api.ts
RitikJaiswal75: Because I changed it to getProgress a common provider tag

PR #764, opened 2023-07-23
refactor : using msw instead of jest.mock of useUserData hook in tests.

Shah-Arsalan:
### Dev Tested:
- [ ] Yes

test-card

test-taskdetails

test-tasklevel

3 files needed change in total wrt to employing msw


__tests__/Unit/Components/Tasks/Card.test.tsx
```diff
@@ -124,7 +122,8 @@ describe('Task card', () => {
         expect(queryByTestId('task-card')).toBeInTheDocument();
     });
 
-    test('should show edit button when ALT key is long pressed', () => {
+    test('should show edit button when ALT key is long pressed', async () => {
+        server.use(...adminUserHandler);
```
bharati-21, line 126: Why are we using aminUserHandler here when we are not doing any testing related to it?

Shah-Arsalan: Beacuse the code is written as `{isUserAuthorized && showEditButton && <EditButton />}` . So to get the isUserAuthorized values as true we need to use adminUserHandler

PR #787, opened 2023-08-01
FIX: remove FF from tasks page

sahsisunny:
### Description:
This PR removes the feature flag that controls the **Intersection Observer** on the tasks page, as it has been thoroughly tested and is ready to be released to all users without the need for the feature flag.

### Dev Tested:
- [x] Yes

### Test Coverage
image
image
image


src/components/tasks/TasksContent.tsx
heyrandhir, line 1: This file seems to be doing something else as well apart from removing the feature flag 🤔

__tests__/Unit/pages/Tasks/index.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    query: {
+        dev: string;
+    };
+    AppTree: AppTreeType;
+};
+
+jest.mock('next/router', () => ({
+    useRouter: () => ({
+        query: {
+            dev: 'true',
```
heyrandhir, line 19: why are we adding the feature flag here dev true if our intension of the PR is to remove it. Also why this new test. we must simply remove the flag from the older tests.

heyrandhir (changes requested):

src/components/tasks/TasksContent.tsx
sahsisunny: Yes removing the old component.

__tests__/Unit/pages/Tasks/index.test.tsx
sahsisunny: It is not part of the feature flag it's a part of SSR. I keeping it because if remove this getInitialProps then the task page calls twice first one for the DEFAULT tab and second one for the SELECTED tab.

heyrandhir (approved):

vinit717 (approved):
First backend has to remove the dev flag and then ask for a merge

heyrandhir (approved):

vinit717 (approved):

Issue #766, opened 2023-07-25
Add search feature for task-title in task-dependency


vinit717:
**Description**

- For a good user experience we have to filter out tasks based on search term so that we can find the task more easily

PR #790, opened 2023-08-03
Add search based filtering for task title

vinit717:
### Issue:

-  https://github.com/Real-Dev-Squad/website-status/issues/766

### Description:
Implement a task title search functionality to add the task in the dependsOn field

### Dev Tested:
- [x] Yes

https://github.com/Real-Dev-Squad/website-status/assets/111434418/c7f3497b-596a-4146-9cf3-9e22e5670227

image
image
image


src/components/taskDetails/taskDependency/index.tsx
```diff
@@ -1,36 +1,114 @@
...
+    ) => void;
+}> = ({ searchTerm, handleChange }) => (
+    <>
+        <Textarea
+            name="dependsOn"
+            value={searchTerm}
+            onChange={handleChange}
+            testId="dependency-textarea"
+        />
+    </>
```
sahsisunny, line 27:
```suggestion
        <Textarea
            name="dependsOn"
            value={searchTerm}
            onChange={handleChange}
            testId="dependency-textarea"
        />
```

__tests__/Unit/Components/Tasks/TaskDependency.test.tsx
```diff
@@ -1,57 +1,99 @@
...
+            },
+            isLoading: false,
+            isError: false,
+            refetch: jest.fn(),
+        });
+
+        const { container } = renderWithRouter(
             <Provider store={store()}>
                 <TaskDependency
+                    taskDependencyIds={[]} // Provide an empty array or relevant data
```
sahsisunny, line 43:
```suggestion
                    taskDependencyIds={[]}
```

vinit717:
Testing an internal state variable very much goes against the philosophy behind react-testing-library. That lib is focused on the user, and what the user can see. The user has no concept of a state variable. Instead of testing that, think of the changes the user would see, and test for that. How is the UI updated? What different markup or styling is displayed? Consider how to test from the user's perspective
https://testing-library.com/docs/guiding-principles/
https://testing-library.com/docs/queries/about/#priority

src/components/taskDetails/taskDependency/index.tsx
vinit717: done

__tests__/Unit/Components/Tasks/TaskDependency.test.tsx
vinit717: done

sahsisunny (approved):

src/components/taskDetails/taskDependency/index.tsx
```diff
@@ -1,36 +1,115 @@
...
+    ) => void;
+}> = ({ searchTerm, handleChange }) => (
+    <Textarea
+        name="dependsOn"
+        value={searchTerm}
+        onChange={handleChange}
+        testId="dependency-textarea"
+        placeholder="Search for task"
+    />
+);
```
Pratiyushkumar, line 27: NIT -Please make this a sperate component

src/components/taskDetails/taskDependency/index.tsx
```diff
@@ -1,36 +1,115 @@
...
+                    data-testid="taskCheckbox"
+                    type="checkbox"
+                    checked={selectedTasks.some((t) => t.id === task.id)}
+                    onChange={() => handleSelectTask(task)}
+                />
+                {task.title}
+            </div>
+        ))}
+    </div>
+);
```
Pratiyushkumar, line 47: NIT - this too...

src/components/taskDetails/taskDependency/index.tsx
vinit717: As it is not usable anywhere so it is unnecessary to create separate file for this

vinit717:
> no dependencies before the loading... is rendered



> After clicking save the name/description does not get updated immediately, the transition from previous to updated name/description is not smooth and also the dependencies shows - no dependencies before the loading... is rendered


https://github.com/Real-Dev-Squad/website-status/assets/111434418/87791780-5353-46b6-9060-defbb411221e
As for saving for name/description, it is an API issue which is a patch request taking some time to update the request and for no dependencies, it is because it first takes the id of the task then it makes a call no it neither fetching nor loading state is their

apurvkhare:
@vinit717 APIs will always take time, even if you optimize. There are two ways to handle this.
1. Optimistic update - update the UI immediately and rollback if the API fails
2. Loading - skeleton or spinner

And regarding dependencies what I meant is that when you do add some dependency and try to save - before it shows loading it shows 'no dependencies' briefly. If this is getting fixed by the above mentioned bugs then fine.

Pratiyushkumar (approved):

PR #823, opened 2023-08-16
Fix the editable mode of task-card for superuser 

AnjaliDeshwani:
### Issue:
Editable mode of task-card for superuser -
1. Task title can not be edited, on clicking title it redirects super user to task details page.
2. There is no icon button to come out of editable mode.
3. On selecting user as a assignee, user's profile picture doesn't reflect correctly corresponding to the assigned user.
4. There is no spinner/checkmark to indicate the user that data is saving/saved.

### Description:

- Changes are made to fix all the above issues. For superuser in editable mode, text area for title is shown. User can still navigate to the task details page on click of title by coming out of editable mode.
- Separate(X) icon is shown, on click of which user can come out of editable mode.
- In Assigned to field, on selecting user from the suggestion list, profile picture also gets reflected correctly.
- In all the editable field spinner is shown when data is being saved, and, once successfully saved checkmark is shown. In case of error, toast is shown.

### Dev Tested:
- [x] Yes

image

image

**Before Change -**

https://github.com/Real-Dev-Squad/website-status/assets/30263070/189ae06b-d27a-433f-b22e-6ba6049d6168

**After Change -**

https://github.com/Real-Dev-Squad/website-status/assets/30263070/6f7b49bf-c15b-4fcc-9069-32f27192a3c2

**When there is error while saving data -**

https://github.com/Real-Dev-Squad/website-status/assets/30263070/acb4e06b-8aaf-403b-9c37-ec2a489d585d

**Test Coverage Before**

TestCoverage_Before

**Test Coverage After**

TestCoverage_After


iamitprakash (changes requested):
in Edit mode Text font looks different, can you align that

AnjaliDeshwani:
> in Edit mode Text font looks different, can you align that

It's fixed now.

src/components/tasks/card/SavedCheckmark.tsx
```diff
@@ -0,0 +1,16 @@
+import Image from 'next/image';
+import classNames from '@/components/tasks/card/card.module.scss';
+export const SavedCheckmark = () => (
+    <div
+        className={`${classNames.checkmark} ${classNames.selfAlignEnd}`}
+        data-testid="checkmark"
+    >
+        <Image
```
iamitprakash, line 8: can we have smaller 'alt' text like "not found" and onerror attribute as well

src/components/tasks/card/TaskStatusEditMode.tsx
```diff
@@ -1,42 +1,76 @@
...
 const beautifyStatus = (status: string) => status.split('_').join(' ');
 const taskStatus = Object.entries(BACKEND_TASK_STATUS);
 
-const TaskStatusEditMode = ({ task, updateTask }: Props) => {
+const TaskStatusEditMode = ({ task, setEditedTaskDetails }: Props) => {
+    const { ERROR } = ToastTypes;
+    const [saveStatus, setSaveStatus] = useState('');
+    const [updateTask] = useUpdateTaskMutation();
+
     const onChangeUpdateTaskStatus = ({
```
iamitprakash, line 25: onChangeUpdateTaskStatus function. The function is supposed to update the task status and set the saveStatus variable to SAVED if the update is successful. However, the function is not checking the response from the updateTask mutation. If the mutation fails, the saveStatus variable will not be set to SAVED and the user will not be notified of the error.

src/components/tasks/card/TaskStatusEditMode.tsx
```diff
@@ -1,42 +1,76 @@
...
+            })
+            .finally(() => {
+                setTimeout(() => {
+                    setSaveStatus('');
+                }, 3000);
+            });
     };
 
     // TODO: remove this after fixing the card beautify status
     const defaultStatus = task.status.toUpperCase().split(' ').join('_');
```
iamitprakash, line 53: in he defaultStatus variable is supposed to be the default value for the status select element. However, the code is converting the task's status to uppercase and then splitting it on spaces. This means that if the task's status is "In Progress", the defaultStatus variable will be set to "IN_PROGRESS". This is not the correct value for the status select element.

src/components/tasks/card/index.tsx
```diff
@@ -354,6 +425,88 @@ const Card: FC<CardProps> = ({
         };
     };
 
+    const onCancelEditRoute = () => {
+        setIsEditMode(false);
+    };
+
+    const CancelEditButton = () => (
```
iamitprakash, line 432: in the tabIndex prop of the Image component. The tabIndex prop is used to control whether the element is focusable or not. A focusable element can be selected with the keyboard and can receive keyboard events. The tabIndex prop should be set to 0 for elements that should be focusable.

In the code you provided, the tabIndex prop of the Image component is set to 0. This means that the Image component is focusable and can receive keyboard events. However, the Image component does not have any functionality that can be controlled with the keyboard. This means that the tabIndex prop of the Image component should be set to -1 to make it non-focusable.

src/components/tasks/card/index.tsx
```diff
@@ -354,6 +425,88 @@ const Card: FC<CardProps> = ({
...
+                setTimeout(() => {
+                    setEditedTaskDetails((prev) => ({
+                        ...prev,
+                        savingText: '',
+                    }));
+                }, 3000);
+            });
+    };
+
+    const debounceTitle = (func: (value: string) => void, delay: number) => {
```
iamitprakash, line 480: in the clearTimeout call. The clearTimeout call is used to cancel a timeout that has already been scheduled. However, the timerIdRef.current variable may be undefined if the debounceTitle function has not been called yet. In this case, the clearTimeout call will throw an error.

src/components/tasks/card/index.tsx
```diff
@@ -354,6 +425,88 @@ const Card: FC<CardProps> = ({
...
+                    savingText: '',
+                }));
+            }
+            timerIdRef.current = setTimeout(() => {
+                func(value);
+            }, delay);
+        };
+    };
+
+    const debouncedHandleTitleChange = debounceTitle(handleTitleChange, 3000);
```
iamitprakash, line 496: inputHandler function is supposed to update the task title and call the debouncedHandleTitleChange function. However, the inputHandler function is not checking to make sure that the value of the input element is not empty. This means that if the user enters an empty value into the input element, the debouncedHandleTitleChange function will not be called and the task title will not be updated.

iamitprakash (changes requested):

vinit717:
There is toast for success or error why you don't use that as previously we are using that

vinit717:
Why not toast for success also? why we are creating different logic for updates as we have a loader, and success toast already

AnjaliDeshwani:
> Why not toast for success also? why we are creating different logic for updates as we have a loader, and success toast already

@vinit717 - It was the requirement told to me by Ankush, as per that we need to show spinner besides editable field when it's being getting saved, and once saved we need to show the checkmark(for few seconds) to indicate that data is being saved.

For error, I am showing toast. If you have any good suggestion to show or make it differently, we can discuss that. Let me know if we need to discuss this.

src/components/tasks/card/SavedCheckmark.tsx
AnjaliDeshwani: updated alt text. For onError I need to check what should be the fallback image for this.

src/components/tasks/card/TaskStatusEditMode.tsx
AnjaliDeshwani: If the mutation fails, it will go inside catch block and will show the toast to the user with the error message. Function is not checking the response from the updateTask mutation inside “then” block because response will be null if mutation is successful.
Please correct me, if my understanding is wrong.

src/components/tasks/card/index.tsx
AnjaliDeshwani: Thanks for pointing it out. I have updated tabIndex prop for the Image to -1. 
As this image can be clicked that’s why I added tabIndex of 0 previously.

src/components/tasks/card/index.tsx
AnjaliDeshwani: clearTimeout will only be called if timerId.current has some value, if it’s undefined, it won’t be called. I have already added if condition to check this “if (timerIdRef.current)”. Let me know if I am understanding it correctly.

src/components/tasks/card/index.tsx
AnjaliDeshwani: I have fixed it now, if user tries to enter the empty value then it will notify the user showing toast that “Title is not allowed to be empty”.

AnjaliDeshwani:
> is this behind feature flag

No, It is not behind feature flag. Will be visible to all users who have the role assigned as superuser.

iamitprakash:
> > is this behind feature flag
> 
> No, It is not behind feature flag. Will be visible to all users in editmode.

please put this under feature flag

iamitprakash (dismissed):

src/components/tasks/card/ShowError.tsx
```diff
@@ -0,0 +1,7 @@
+import Image from 'next/image';
+import classNames from '@/components/tasks/card/card.module.scss';
+export const ShowError = () => (
+    <div className={classNames.selfAlignEnd} data-testid="checkmark">
+        <Image src="/error.png" alt="data saved" width={20} height={20} />
```
vvaibhavdesai, line 5: the alt tag seems a little odd here. can be something else

src/components/tasks/card/SavedCheckmark.tsx
```diff
@@ -0,0 +1,7 @@
+import Image from 'next/image';
+import classNames from '@/components/tasks/card/card.module.scss';
+export const SavedCheckmark = () => (
```
vvaibhavdesai, line 3: add a line gap  on the top after import

src/components/tasks/card/ShowError.tsx
```diff
@@ -0,0 +1,7 @@
+import Image from 'next/image';
+import classNames from '@/components/tasks/card/card.module.scss';
+export const ShowError = () => (
```
vvaibhavdesai, line 3: add a line gap  on the top after import

src/components/tasks/card/SavedCheckmark.tsx
```diff
@@ -0,0 +1,7 @@
...
+    <div className={classNames.selfAlignEnd} data-testid="checkmark">
+        <Image src="/check.png" alt="data saved" width={20} height={20} />
```
vvaibhavdesai, line 5: alt tag here as well seems odd

src/components/tasks/card/SmallSpinner.tsx
```diff
@@ -0,0 +1,7 @@
+import classNames from '@/components/tasks/card/card.module.scss';
+export const SmallSpinner = () => (
```
vvaibhavdesai, line 2: skip a line after imports

src/components/tasks/card/TaskStatusEditMode.tsx
```diff
@@ -1,42 +1,84 @@
...
+    const [saveStatus, setSaveStatus] = useState('');
+    const [updateTask] = useUpdateTaskMutation();
+
     const onChangeUpdateTaskStatus = ({
         target: { value },
     }: React.ChangeEvent<HTMLSelectElement>) => {
-        updateTask(task.id, {
-            status: value,
+        setSaveStatus(PENDING);
+        setEditedTaskDetails((prev: any) => ({ ...prev, status: value }));
```
vvaibhavdesai, line 32: any reasons specific reasons to use any ?

src/components/tasks/card/TaskStatusEditMode.tsx
```diff
@@ -1,42 +1,84 @@
...
+                    {taskStatus.map(([name, status]) => (
+                        <option key={status} value={status}>
+                            {beautifyStatus(name)}
+                        </option>
+                    ))}
+                </select>
+            </label>
+            {dev === 'true' && (
+                <>
+                    {saveStatus === PENDING && <SmallSpinner />}
```
vvaibhavdesai, line 76: maybe take this out in a if else statement and make it a component ?

src/components/tasks/constants.ts
```diff
@@ -1,3 +1,6 @@
 export const INITIAL_TASKS_LIMIT = 3;
 export const ADD_MORE_TASKS_LIMIT = 5;
 export const SEE_MORE = 'See More';
+export const PENDING = 'Pending';
```
vvaibhavdesai, line 4: would suggest using the string in all caps until we have a very specific reason of doing this ?

vvaibhavdesai (commented):
Overall nicely done great work 🚀

public/cancel.png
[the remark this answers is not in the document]
AnjaliDeshwani: I tried to convert these png images to svg but it's not coming out exactly same.
Will replace these if I find svgs for these images.

src/components/tasks/card/ShowError.tsx
AnjaliDeshwani: Added.

src/components/tasks/card/SavedCheckmark.tsx
AnjaliDeshwani: I have Updated it now.

src/components/tasks/card/TaskStatusEditMode.tsx
AnjaliDeshwani: Thanks for pointing it out, I have now updated it with correct type.

src/components/tasks/constants.ts
AnjaliDeshwani: I have updated it now with all caps. Missed to follow this convention.

sahsisunny:
Can you add the test coverage stats?

src/components/tasks/card/TaskStatusEditMode.tsx
AnjaliDeshwani: Great suggestion, I have made the separate component for it now.

src/components/tasks/card/ShowError.tsx
AnjaliDeshwani: I have Updated it now.

AnjaliDeshwani:
> Can you add the test coverage stats?

@sahsisunny - I have added few test cases and updated the coverage please check. Thank you.

src/components/tasks/card/SmallSpinner.tsx
AnjaliDeshwani: Updated at all the places

Achintya-Chatterjee (dismissed):
LGTM

Pratiyushkumar (dismissed):

Pratiyushkumar (approved):

Achintya-Chatterjee (dismissed):

AnjaliDeshwani:
Closing this PR as because of rebase and reverting changes again it's facing some conflicting issues while someone approves the PR.

Opened the new PR - https://github.com/Real-Dev-Squad/website-status/pull/839

PR #868, opened 2023-09-13
Disable "Task-Request" button when the task is not available.

### LINKED ISSUE #862: [TaskDetails] - The task request button is enabled, even when task is assigned
status-site-issue2

As we can see, the task is assigned even then the task request button is enabled. This should not be the case, the REQUEST FOR TASK button should only be enable when there is no Assignee.

What we need to do:

- Enable REQUEST FOR TASK button only when the task has no assignee.

0xatulpatil:
fixes  #862 

### Issue:
The "Request for task" button is enabled even when the task is assigned to someone else. The REQUEST FOR TASK button should only be enabled when there is no assignee.

### Dev Tested:
- [x] Yes

image
image

### Test Stats:
image


src/components/taskDetails/index.tsx
```diff
@@ -352,6 +352,12 @@ const TaskDetails: FC<Props> = ({ taskID }) => {
                                             data-testid="request-task-button"
                                             className={classNames.button}
                                             onClick={taskRequestHandle}
+                                            disabled={
+                                                !Object.is(
+                                                    taskDetailsData?.assignee,
+                                                    null
```
Pratiyushkumar, line 358:
```suggestion
                                                !Object.is(
                                                    taskDetailsData?.assignee,
                                                    null
```
`const isTaskDeatilsData =  !Object.is(taskDetailsData?.assignee,null)`
and then use this variable as disabled value

src/components/taskDetails/task-details.module.scss
```diff
@@ -40,6 +40,11 @@ $off-white: #f0f0f0;
         background-color: $theme-blue;
         color: $white;
     }
+    &:disabled {
+        color: #616161;
```
Pratiyushkumar, line 44: use color variable name for color

Pratiyushkumar (changes requested):
put test coverage as well

src/components/taskDetails/index.tsx
```diff
@@ -352,6 +352,12 @@ const TaskDetails: FC<Props> = ({ taskID }) => {
                                             data-testid="request-task-button"
                                             className={classNames.button}
                                             onClick={taskRequestHandle}
+                                            disabled={
+                                                !Object.is(
```
TanishqSingla, line 356: This equality check will only check for the object to be null or not, in case it is undefined which we might get in the case when assignee is not present, or any other falsy value this equality check will return `true`

TanishqSingla (changes requested):

__mocks__/handlers/task-details.handler.ts
```diff
@@ -70,6 +70,37 @@ const taskDetailsHandler = [
             })
         );
     }),
+    rest.get(`${URL}/tasks/6KhcLU3yr45dzjQIVm0l/details`, (_, res, ctx) => {
```
TanishqSingla, line 73: You can use wild cards and slugs, that way we can add different ids and different behavior in one handler itself.
Check the path params section in this doc https://mswjs.io/docs/basics/request-matching

TanishqSingla (changes requested):

--hostname
Pratiyushkumar, line 1: please remove this file, I dont think it is required

Issue #867, opened 2023-09-13
Fix: Disable Task request button on assigned tasks.


kapillondhe:
Description: 
The task request button should not be enabled for assigned tasks as highlighted in the image below.

Image:
![image](https://github.com/Real-Dev-Squad/website-status/assets/41846637/016fd6c1-d6a1-41d0-81b8-ddd7128ccbd1)

kapillondhe:
Alternate issue #862 , so closing this issue.

PR #871, opened 2023-09-15
Fix: Disable Task request button on assigned tasks.

kapillondhe:
### Issue:
#867 
The "Request for task" button should not be enabled for assigned tasks as highlighted in the image below.

Image:
image

### Description:
Disabled the "Request for task" button when the task status is "ASSIGNED".
Added tests to check if the button is enabled or disabled for different task status.

No

### Dev Tested:
- [✓ ] Yes

image

[screen-capture (1).webm](https://github.com/Real-Dev-Squad/website-status/assets/41846637/abf43514-2cd6-44a2-8718-8de70c46b5e7)

### Follow-up Issues (if any)
No

### Test Stat
Before:
image

After:
image

### DIFF
--- a/__mocks__/handlers/task-details.handler.ts
+++ b/__mocks__/handlers/task-details.handler.ts
@@ -70,6 +70,42 @@ const taskDetailsHandler = [
             })
         );
     }),
+    rest.get(`${URL}/tasks/6KhcLU3yr45dzjQIVm0l/details`, (_, res, ctx) => {
+        return res(
+            ctx.status(200),
+            ctx.json({
+                message: 'task returned successfully',
+                taskData: {
+                    id: '6KhcLU3yr45dzjQIVm0l',
+                    isNoteworthy: true,
+                    lossRate: {
+                        dinero: 0,
+                        neelam: 0,
+                    },
+                    endsOn: 1618790410,
+                    title: 'test 1 for drag and drop',
+                    status: 'in_progress',
+                    assignee: 'ankur',
+                    links: ['null'],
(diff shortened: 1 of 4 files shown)


src/components/taskDetails/index.tsx
```diff
@@ -352,6 +352,10 @@ const TaskDetails: FC<Props> = ({ taskID }) => {
                                             data-testid="request-task-button"
                                             className={classNames.button}
                                             onClick={taskRequestHandle}
+                                            disabled={
+                                                taskDetailsData?.status.toUpperCase() ===
+                                                'ASSIGNED'
+                                            }
```
vinit717, line 358: instead of status go for if task details have an assignee or not because if status is in_progress then you can request for task

__mocks__/handlers/task-details.handler.ts
```diff
@@ -70,6 +70,42 @@ const taskDetailsHandler = [
...
+                    },
+                    github: {
+                        issue: {
+                            html_url:'https://github.com/sample-org/sample-repo/issues/000'
+                        }
+                    },
+                },
+            })
+        );
+    }),
```
vinit717, line 108: I think we have already mock handler for this

src/components/taskDetails/index.tsx
sahsisunny: Yes you can use UNASSIGNED or AVAILABLE status key for disabling the **Request Task** button.

kapillondhe:
Alternate PR #868 by Atul. So closing this PR.

Issue #899, opened 2023-09-28
Fix the invalid date bug


Maheima:
When the startedOn date isn't present display N/A instead of invalid date

PR #906, opened 2023-09-30
Fix the invalid date bug

Maheima:
### Issue:
Currently, on the cards where startedOn isn't available, it shows invalid date. We need to change it to N/A instead
Fixes: #899 

### Description:
Wrote a function that gets from startedon date only when we get started on date of that task from the API

### Files changed:
__tests__/Unit/Components/Tasks/TaskDetails.test.tsx -> Added tests cases
src/components/taskDetails/index.tsx ->  Fixed the NA to N/A

### Dev Tested:
- [x] Yes

Video of the change:
https://www.loom.com/share/031d00dd92c34ffa8cb8cb0bb53fa978?sid=ec61421a-c7ee-44b8-b7f3-cdc1665b12cd

### Test coverage:

Before:

image

After:

image

### DIFF
--- a/src/components/taskDetails/index.tsx
+++ b/src/components/taskDetails/index.tsx
@@ -312,15 +312,23 @@ const TaskDetails: FC<Props> = ({ taskID }) => {
                             >
                                 <Details
                                     detailType={'Started On'}
-                                    value={getTaskEstimates(
+                                    value={
                                         taskDetailsData?.startedOn
-                                    )}
+                                            ? convertTimeStamp(
+                                                  taskDetailsData.startedOn
+                                              )
+                                            : 'N/A'
+                                    }
                                 />
                                 <Details
                                     detailType={'Ends On'}
-                                    value={getTaskEstimates(
+                                    value={
                                         taskDetailsData?.endsOn
-                                    )}
+                                            ? convertTimeStamp(
+                                                  taskDetailsData.endsOn
+                                              )
+                                            : 'TBD'
+                                    }
                                 />
                             </TaskContainer>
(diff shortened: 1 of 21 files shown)


src/components/taskDetails/index.tsx
```diff
@@ -158,8 +158,12 @@ const TaskDetails: FC<Props> = ({ taskID }) => {
         }
     }
 
-    function getTaskEstimates(timestamp: number | undefined) {
-        return timestamp ? convertTimeStamp(timestamp) : 'NA';
+    function getStartedOn(timestamp: number | undefined) {
+        return timestamp ? convertTimeStamp(timestamp) : 'N/A';
```
iamitprakash, line 162: what if timestamp undefined ...?

src/components/taskDetails/index.tsx
```diff
@@ -158,8 +158,12 @@ const TaskDetails: FC<Props> = ({ taskID }) => {
...
+    }
+
+    function getEndsOn(timestamp: number | undefined) {
+        return timestamp ? convertTimeStamp(timestamp) : 'TBD';
```
iamitprakash, line 166: what if timestamp is undefined ...?

iamitprakash (changes requested):

src/components/taskDetails/index.tsx
Maheima: But that is what I'm checking for . If it is undefined, it goes to the else condition i.e., N/A

iamitprakash (approved):

Achintya-Chatterjee (approved):
LGTM

PR #930, opened 2023-10-11
[Bug] Status Site - Task cards are breaking UI due to title text getting truncated without text wrap

aneenajohn:
### Description:
- On [Tasks Page](https://status.realdevsquad.com/tasks?q=status%3Aall), when the title text doesn't fit in one line, the card title has white-space set to no-wrap causing the UI to break.

- PFB the screenshots
image
image

- With white-space set to normal prevents UI break as below:
image

### Anthing you would like to inform the reviewer about:
- CSS Fixes: When the title text doesn't fit, it gets truncated and instead of setting no-wrap to white-space we let the text wrap preserving spaces by setting white-space to pre-wrap

### Dev Tested:
- [✅] Yes
- test ss

https://github.com/Real-Dev-Squad/website-status/assets/75213145/674aa82c-4495-4c3f-8b66-ffe45d02a644

### Follow-up Issues (if any)
NA

### DIFF
--- a/src/components/tasks/card/index.tsx
+++ b/src/components/tasks/card/index.tsx
@@ -520,7 +520,7 @@ const Card: FC<CardProps> = ({
                         taskId={cardDetails.id}
                     >
                         <span
-                            className={classNames.cardTitle}
+                            className={classNames.cardTitleText}
                             contentEditable={isEditable}
                             onKeyDown={(e) => handleChange(e, 'title')}
                             role="button"
--- a/src/components/tasks/card/card.module.scss
+++ b/src/components/tasks/card/card.module.scss
@@ -79,18 +79,29 @@ $progressTextAnimation: #9babb8;
     font-size: 1.6rem;
     overflow: hidden;
     text-overflow: ellipsis;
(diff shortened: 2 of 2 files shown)


src/components/tasks/card/index.tsx
```diff
@@ -520,7 +520,7 @@ const Card: FC<CardProps> = ({
                         taskId={cardDetails.id}
                     >
                         <span
-                            className={classNames.cardTitle}
+                            className={classNames.cardTitleText}
```
sahsisunny, line 523: Can we remove the CSS for the `cardTitle` class if it's not being used?

aneenajohn: @sahsisunny, `cardTitle` class is being used on the parent component `ConditionalLinkWrapper` over the `span` title element.

sahsisunny: Okay

sahsisunny (approved):

shubhamsigdar1 (approved):

Achintya-Chatterjee (approved):

PR #933, opened 2023-10-13
add UTC time format on idle page

anishpawaskar:
### Description:
- The idle page where we show idle users and their time/day for how long they have been idle. So to show time/day we are not using UTC time format.
- So i have added UTC time format there.

### Anthing you would like to inform the reviewer about:
- Didn't write any test cause it is not require and test are already written for card component.
 
### Dev Tested:
- [x] Yes

Image proof :-
Idle Users _ Status Real Dev Squad _ Status Real Dev Squad - Google Chrome 14-10-2023 02_50_09

### DIFF
--- a/src/helperFunctions/getIdleSinceText.ts
+++ b/src/helperFunctions/getIdleSinceText.ts
@@ -5,11 +5,12 @@ import {
 
 const getIdleSinceText = (idleSince: string) => {
     const presentDate = new Date();
+    const idleSinceDate = new Date(idleSince);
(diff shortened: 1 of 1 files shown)


iamitprakash (approved):

Achintya-Chatterjee (commented):
@anishpawaskar add a working proof and also tests the stats

Achintya-Chatterjee (approved):

j24m (approved):

PR #1060, opened 2023-12-26
fix: dropdown link for my site

sahsisunny:
### Issue: 
- #1064

### Description:

This PR addresses the issue #1064, which reported a problem where clicking on the 'Status' dropdown would open the user profile instead of displaying the expected status options. The root cause of this behavior was identified and rectified in the following manner:

### Dev Tested:
- [x] Yes


satyam73 (approved):
LGTM

vinit717 (approved):

PR #1308, opened 2025-01-04
Update Progress Fetching Logic to Include `dev` Query Parameter

### LINKED ISSUE #1307: Add dev Parameter to Fetch Task Progress Conditionally
### Issue Description

The frontend needs to handle the `dev=true` query parameter to fetch the `userData` field in the progress response. This functionality depends on a backend PR that is currently under review. This PR is a draft, and once the backend PR is merged, it can be finalized and marked for review.
### Expected Behavior

When `dev=true` is passed in the query, the API should return the `userData` field along with the progress data. The frontend should send this parameter in the request.
### Current Behavior

Currently, the frontend does not pass the `dev=true` query parameter, and as a result, the `userData` field is not included in the progress response.

### Severity/Priority
 - [ ] Critical
 - [ ] High
 - [x] Medium
 - [ ] Low

### Checklist

- [x] I have read and followed the project's code of conduct.
- [x] I have searched for similar issues before creating this one.
- [x] I have provided all the necessary information to understand and reproduce the issue.
- [x] I am willing to contribute to the resolution of this issue.

AnujChhikara:
Date: 4 Jan, 2025

Developer Name: @AnujChhikara 

---

## Issue Ticket Number

- Closes #1307
## Description
This PR introduces support for the dev query parameter when fetching task progress data. The `dev=true` parameter is used to conditionally fetch the `userData` field along with the progress data. This change ensures that the frontend correctly handles the parameter when requesting task progress details from the backend.

### Documentation Updated?

- [ ] Yes
- [x] No

### Under Feature Flag

- [x] Yes
- [ ] No

### Database Changes

- [ ] Yes
- [x] No

### Breaking Changes

- [ ] Yes
- [x] No

### Development Tested?

- [x] Yes
- [ ] No

## Screenshots
<details>
<summary>Screenshot 1</summary>

ShareX_LE4OZJqbos

</details>

## Test Coverage
<details>
<summary>Screenshot 1</summary>

image

</details>


src/components/taskDetails/index.tsx
```diff
@@ -64,6 +64,8 @@ type Props = {
 
 const TaskDetails: FC<Props> = ({ taskID }) => {
     const router = useRouter();
+    const { dev } = router.query;
+    const isDev = dev === 'true';
```
pankajjs, line 68:
```suggestion
    const { dev } = router.query;
    const isDev = dev === 'true';
```
Can we do it like this?
```
const isDev = router.query.dev === "true";
```
because dev parameter is only being used in line 67 and 68.

__tests__/Unit/hooks/progressesApi.test.tsx
```diff
@@ -43,4 +43,31 @@ describe('useGetProgressDetails', () => {
         expect(taskProgressResponse).not.toBeUndefined();
         expect(taskProgressResponse).toEqual(mockGetTaskProgress);
     });
+    test('Returns Task Progress with userData in response if dev is true', async () => {
+        const { result, waitForNextUpdate } = renderHook(
```
pankajjs, line 47: Do we have any test when dev is false and only progress data is returned?

AnujChhikara: Currently not specially for dev=false . i have added that test for that you can check. Thank you

src/components/taskDetails/index.tsx
AnujChhikara: resolved

__mocks__/db/progresses.ts
```diff
@@ -102,43 +102,8 @@ export const mockGetTaskProgress = {
...
-                    url: 'https://res.cloudinary.com/realdevsquad/image/upload/v1692058952/profile/XAF7rSUvk4p0d098qWYS/me40uk7taytbjaa67mhe.jpg',
-                },
-                github_display_name: 'Ankush Dharkar',
-                company_name: 'Amazon',
-                github_id: 'ankushdharkar',
-                designation: 'SDE',
-                status: 'idle',
-                username: 'ankush',
-            },
-            taskId: 'OxYqJgf6Tyl90uci1mzs',
```
pankajjs, line 141: why did we remove it?

AnujChhikara: Because in the mock data for progresses we don't have any object without `userData` . Either I have to remove data from object of create a separate object for it

pankajjs: We need two progress data one without userData and one with userData right.
So instead of removing this, we can use it where dev = true.
For dev = false, we can do it like this progress.userData = undefined when passing as mock data.

AnujChhikara: I can create two different mock data one with `userData` and one without `userData` and then based on dev value pass them as mock data. Is this approach right?

pankajjs: Why two objects? We can use one with userData for dev = true and for dev = false, just modify the same object and make the userData undefined, `progress.userData = undefined` and pass the progress data.

AnujChhikara: In my current solution, we have three progress objects, all of which include userData. To handle the case where userData is missing, I modified one of these objects and removed userData field from that. This allows me to use it for scenarios where dev = false.

pankajjs: Instead of modifying the whole data, Can't we just make the userData undefined before passing it as mock?

AnujChhikara: @pankajjs can you please check it now.

pankajjs (approved):
LGTM

fakhruddinkw (approved):
LGTN

